PR #1, opened 2023-01-12
Fix initial addon output

simonihmig:
The blueprint output has some issues that need to be fixed upstream. Namely the use of deprecated `@types/ember__test-helpers` (that somehow breaks the rollup-ts plugin), and missing `.gitignore` file.


NullVoxPopuli (approved):
💯

PR #14, opened 2023-01-13
Implement basic component API

### LINKED ISSUE #3: Implement basic component API
Implement Form and its yielded components, accept data and submit.

Basically a simple form like this should work:

```hbs
<HeadlessForm
  @data={{this.data}}
  @onSubmit={{this.doSomething}}
  as
  |form|
>
  <form.field @name="name" as |field|>
    <field.label>Name</field.label>
    <field.input @type="text" />
  </form.field>

  <form.submitButton>Submit</form.submitButton>
</HeadlessForm>
```

simonihmig:
Support the most basic component API:
* `<HeadlessForm>` as the _only_ public component
* yielding `field`, which in turn yields `label` and `input` components, as well as `value` and `setValue` for custom markup
* accepts (typed!) `@data`, passed all the way down to `input`
* calls `@onSubmit`

```hbs
<HeadlessForm
  @data={{this.data}}
  @onSubmit={{this.doSomething}}
  as
  |form|
>
  <form.field @name="name" as |field|>
    <field.label>Name</field.label>
    <field.input @type="text" />
  </form.field>

  <button type="submit">Submit</button>
</HeadlessForm>
```
Closes #3


ember-headless-form/src/components/-private/control/input.ts
```diff
@@ -0,0 +1,42 @@
+import Component from '@glimmer/component';
+
+// Possible values for the input type, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types
+// for the sake ok completeness, we list all here, with some commented out that are better handled elsewhere, or not at all...
+export type InputType =
```
NullVoxPopuli, line 5: 🎉

ember-headless-form/src/components/-private/control/input.ts
```diff
@@ -0,0 +1,42 @@
...
+  | 'url'
+  | 'week';
+
+export interface HeadlessFormControlInputComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    value: string | undefined;
+    type?: InputType;
+    fieldId: string;
+  };
```
NullVoxPopuli, line 35: Blocks is missing?
```ts
Blocks: {
  default: [components: {
    label: ...
    input: ...
  }];
}
```

NullVoxPopuli: ha, nevermind, I was looking at the wrong template

ember-headless-form/src/components/-private/field.ts
```diff
@@ -0,0 +1,45 @@
...
+> extends Component<HeadlessFormFieldComponentSignature<DATA>> {
+  LabelComponent = LabelComponent;
+  InputComponent = InputComponent;
+
+  get value(): DATA[keyof DATA] | undefined {
+    return this.args.data[this.args.name];
+  }
+
+  // copy-pasted from https://github.com/emberjs/ember.js/blob/master/packages/@ember/-internals/glimmer/lib/helpers/unique-id.ts
+  // we can remove this once we only support Ember 4.4+!
```
NullVoxPopuli, line 35: anything wrong with only supporting Ember 4.4+?

simonihmig: Hm, good question. No, probably not? 🤷‍♂️ At least for a fresh new addon...

simonihmig: Only supporting Ember 4.4+, and using native `{{unique-id}}`

ember-headless-form/rollup.config.mjs
```diff
@@ -24,7 +24,7 @@ export default {
     // These are the modules that should get reexported into the traditional
     // "app" tree. Things in here should also be in publicEntrypoints above, but
     // not everything in publicEntrypoints necessarily needs to go here.
-    addon.appReexports(['components/**/*.js']),
+    addon.appReexports(['components/*.js']),
```
NullVoxPopuli, line 27: this ended up working?

ember-headless-form/unpublished-development-types/index.d.ts
```diff
@@ -2,12 +2,23 @@
 // These will *not* be published as part of your addon, so be careful that your published code does not rely on them!
 
 import '@glint/environment-ember-loose';
+import type HeadlessFormRegistry from '../src/template-registry';
+import { ComponentLike } from '@glint/template';
```
NullVoxPopuli, line 6:
```suggestion
import type { ComponentLike } from '@glint/template';
```
surprised linting didn't catch this 🤔 
or maybe it doesn't matter since in a .d.ts file, everything is a type?

ember-headless-form/rollup.config.mjs
simonihmig: Yeah, publicEntrypoints still has the fully `**` glob, for the reason discussed yesterday. This just amkes sure our app re-exportsin `package.json` only exposes the root component to the app, when doing the classic string lookups fomr DI rather than imports:

image

So this is unrelated to the Glint issue we discussed, if this was your concern?

test-app/.prettierrc.js
```diff
@@ -1,11 +1,20 @@
 'use strict';
 
 module.exports = {
+  plugins: ['prettier-plugin-ember-template-tag'],
+  singleQuote: true,
+  templateSingleQuote: false,
   overrides: [
     {
-      files: '*.{js,ts}',
+      files: '*.gjs',
```
NullVoxPopuli, line 9: is this not automatic? 🤔 

maybe I missed something the the README for this plugin

simonihmig: Yeah, probably this shouldn't be there. But this was the only way to make the prettier VSCode plugin work, see https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/38

ember-headless-form/unpublished-development-types/index.d.ts
simonihmig: Oh, I think I don't have your eslint-config for the addon (just the blueprint defaults here), only for the test-app (due to template-imports). I can add it here as well...

ember-headless-form/src/components/-private/field.hbs
```diff
@@ -0,0 +1,16 @@
+{{#let (unique-id) (fn @set @name) as |uuid setValue|}}
```
nicolechung, line 1: Question: how come these `field` files are just `index` files inside of the field folder?

ember-headless-form/rollup.config.mjs
NullVoxPopuli: ahh! yup, I was confused. thanks!

.github/workflows/ci.yml
```diff
@@ -62,12 +62,11 @@ jobs:
       fail-fast: false
       matrix:
         try-scenario:
-          - ember-lts-3.28
```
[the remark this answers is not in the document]
simonihmig: Well, this is the very start of everything, so we never supported _anything_ before! 😆
But you are right, we are starting by not supporting Ember 3.28, at least that's the outcome of some back and forth between Preston and me. But we can revisit this decision, if you think there is a good reason to still support Ember 3.28. 

Having Ember 3.28 support is good for _existing_ addons, as that supports apps that still have trouble upgrading to 4. And itÄs quite unfortunate if addons don't have this overlap, and maybe only the newer version of an addon supports Ember 4, but not Ember 3.28. In this way users are forces to upgrade Ember _and_ the addon at the same time.

But none of this applies here, as this is a green field project, and we don't have any existing Ember 3.28 users, so we can just support what is reasonable for us I think...

NullVoxPopuli (approved):
I think this is a great initial implementation. We're still working out some details on Slack, but I see no reason to hold up this PR. nice work!

ember-headless-form/src/components/-private/field.hbs
simonihmig: Not sure what you mean here with `index` files?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -0,0 +1,415 @@
...
+        await click('[data-test-submit]');
+
+        assert.deepEqual(
+          data,
+          { firstName: 'Tony', lastName: 'Ward' },
+          'data is not mutated'
+        );
+
+        assert.true(
+          submitHandler.calledWith({ firstName: 'Nicole', lastName: 'Chung' }),
```
nicolechung, line 376: Hey that's me!

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -0,0 +1,415 @@
...
+          <form.field @name="firstName" as |field|>
+            <field.label>First Name</field.label>
+            <field.input />
+          </form.field>
+        </HeadlessForm>
+      </template>);
+
+      assert.dom('input').hasAttribute(
+        'id',
+        // copied from https://ihateregex.io/expr/uuid/
```
nicolechung, line 161: TIL

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -0,0 +1,415 @@
...
+        </HeadlessForm>
+      </template>);
+
+      assert
+        .dom('[data-test-user-content]')
+        .exists('form field can render user content');
+
+      assert
+        .dom('form > [data-test-user-content]')
+        .exists('does not render anything on its own');
```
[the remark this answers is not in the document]
simonihmig: Yeah, I see how this can be a bit confusing. The selector basically tests that there is no DOM node sitting between `<form>` and our `<div data-test-user-content>`, i.e. the `field` component does not add any markup on its own (which we want to do as little as possible, given were are "headless")

nicolechung (approved):
Looks good! I left a few tiny comments / questions but more out of curiosity

test-app/tests/integration/components/headless-form-test.gts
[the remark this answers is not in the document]
nicolechung: Oh I see! Maybe it can be `field component contains no markup itself` or something then.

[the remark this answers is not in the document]
simonihmig: Yeah, that's a _much_ better phrase, applied that. TY!

test-app/.prettierrc.js
NullVoxPopuli: making editors work is pretty important, I'm in favor

simonihmig: I am adding a comment there to point to that issue, so we can revert that should that get fixed.

ember-headless-form/unpublished-development-types/index.d.ts
simonihmig: I added `@nullvoxpopuli/eslint-configs` also for the addon now, and indeed `--fix` did a couple of changes!

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -0,0 +1,415 @@
...
+
+        assert.dom('input').hasAttribute('type', type, `supports type=${type}`);
+      }
+    });
+  });
+
+  module('data', function () {
+    module('data down', function () {
+      test('data is passed to form controls', async function (assert) {
+        const data = { firstName: 'Tony', lastName: 'Ward' };
```
ynotdraw, line 238: I'm forever now part of ember-headless-forms! 🎉

ynotdraw (approved):
Echoing @NullVoxPopuli's statement, this is an excellent step forward! Thanks a bunch!

PR #15, opened 2023-01-19
Support checkbox control

### LINKED ISSUE #8: Support for `<input type="checkbox">` form control

simonihmig:
Besides adding `field.checkbox`, this refactors a bit the internal types, related to passing only the suitable value type to a control: `field.input` should only handle `string` values, while `field.checkbox` is for `boolean`s.

Closes #8 

- [x] Blocked on merging #14 first!

### DIFF
--- a/ember-headless-form/src/components/-private/control/checkbox.ts
+++ b/ember-headless-form/src/components/-private/control/checkbox.ts
@@ -1,5 +1,4 @@
 import Component from '@glimmer/component';
-import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 
 export interface HeadlessFormControlCheckboxComponentSignature {
--- a/ember-headless-form/src/components/-private/control/checkbox.ts
+++ b/ember-headless-form/src/components/-private/control/checkbox.ts
@@ -1,0 +1,19 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+import { action } from '@ember/object';
+
+export interface HeadlessFormControlCheckboxComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    value: boolean;
+    fieldId: string;
+    setValue: (value: boolean) => void;
+  };
+}
+
+export default class HeadlessFormControlCheckboxComponent extends Component<HeadlessFormControlCheckboxComponentSignature> {
+  @action
+  handleInput(e: Event | InputEvent): void {
+    this.args.setValue((e.target as HTMLInputElement).checked);
+  }
+}
--- a/ember-headless-form/src/components/-private/control/input.hbs
+++ b/ember-headless-form/src/components/-private/control/input.hbs
@@ -1,6 +1,6 @@
 <input
   type={{@type}}
-  value={{this.valueAsString}}
+  value={{@value}}
   id={{@fieldId}}
   ...attributes
   {{on 'input' this.handleInput}}
--- a/ember-headless-form/src/components/-private/control/input.ts
+++ b/ember-headless-form/src/components/-private/control/input.ts
@@ -59,6 +59,7 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
 
   @action
   handleInput(e: Event | InputEvent): void {
-    this.args.setValue((e.target as HTMLInputElement).value);
+    assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
+    this.args.setValue(e.target.value);
   }
 }
--- a/ember-headless-form/src/components/-private/control/input.ts
+++ b/ember-headless-form/src/components/-private/control/input.ts
@@ -28,22 +28,20 @@ export type InputType =
   | 'url'
   | 'week';
 
-export interface HeadlessFormControlInputComponentSignature<VALUE> {
+export interface HeadlessFormControlInputComponentSignature {
   Element: HTMLInputElement;
   Args: {
-    value: VALUE;
+    value: string;
     type?: InputType;
     fieldId: string;
-    setValue: (value: VALUE) => void;
+    setValue: (value: string) => void;
   };
 }
 
-export default class HeadlessFormControlInputComponent<VALUE> extends Component<
-  HeadlessFormControlInputComponentSignature<VALUE>
-> {
+export default class HeadlessFormControlInputComponent extends Component<HeadlessFormControlInputComponentSignature> {
   constructor(
     owner: unknown,
-    args: HeadlessFormControlInputComponentSignature<VALUE>['Args']
+    args: HeadlessFormControlInputComponentSignature['Args']
   ) {
     assert(
       `input component does not support @type="${args.type}" as there is a dedicated component for this. Please use the \`field.${args.type}\` instead!`,
@@ -59,18 +57,8 @@ export default class HeadlessFormControlInputComponent<VALUE> extends Component<
     return this.args.type ?? 'text';
   }
 
-  get valueAsString(): string {
-    assert(
-      `input can only handle string values, but you passed ${typeof this.args
-        .value}`,
-      typeof this.args.value === 'string'
-    );
-
-    return this.args.value;
-  }
-
   @action
   handleInput(e: Event | InputEvent): void {
-    this.args.setValue((e.target as HTMLInputElement).value as VALUE);
+    this.args.setValue((e.target as HTMLInputElement).value);
   }
 }
--- a/ember-headless-form/src/components/-private/control/input.ts
+++ b/ember-headless-form/src/components/-private/control/input.ts
@@ -41,6 +41,20 @@ export interface HeadlessFormControlInputComponentSignature<VALUE> {
 export default class HeadlessFormControlInputComponent<VALUE> extends Component<
   HeadlessFormControlInputComponentSignature<VALUE>
 > {
+  constructor(
+    owner: unknown,
+    args: HeadlessFormControlInputComponentSignature<VALUE>['Args']
+  ) {
+    assert(
+      `input component does not support @type="${args.type}" as there is a dedicated component for this. Please use the \`field.${args.type}\` instead!`,
+      args.type === undefined ||
+        // TS would guard us against using an unsupported `InputType`, but for JS consumers we add a dev-only runtime check here
+        (args.type as string) !== 'checkbox'
+    );
+
+    super(owner, args);
+  }
+
   get type(): InputType {
     return this.args.type ?? 'text';
   }
--- a/ember-headless-form/src/components/-private/field.hbs
+++ b/ember-headless-form/src/components/-private/field.hbs
@@ -8,6 +8,12 @@
         value=this.value
         setValue=setValue
       )
+      checkbox=(component
+        (ensure-safe-component this.CheckboxComponent)
+        fieldId=uuid
+        value=this.value
+        setValue=setValue
+      )
       value=this.value
       id=uuid
       setValue=setValue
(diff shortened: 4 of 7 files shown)


ember-headless-form/src/components/-private/control/checkbox.ts
```diff
@@ -0,0 +1,19 @@
...
+export interface HeadlessFormControlCheckboxComponentSignature {
+  Element: HTMLInputElement;
+  Args: {
+    value: boolean;
+    fieldId: string;
+    setValue: (value: boolean) => void;
+  };
+}
+
+export default class HeadlessFormControlCheckboxComponent extends Component<HeadlessFormControlCheckboxComponentSignature> {
```
NullVoxPopuli, line 14: are these names from the generator? they're very long 😅

simonihmig: No, they are a product of my own creativity! 🤷‍♂️ 😆

Seriously though, even when verbose, they follow the logical (folder) structure, which helps me to keep them consistent...

ynotdraw (approved):

ember-headless-form/src/components/-private/control/input.ts
```diff
@@ -28,35 +28,37 @@ export type InputType =
...
-    return this.args.value;
+    super(owner, args);
+  }
+
+  get type(): InputType {
+    return this.args.type ?? 'text';
   }
 
   @action
   handleInput(e: Event | InputEvent): void {
```
nicolechung, line 61: Suggestion, nit: Instead of a cast here, can we use a custom type guard?

pseudocode, not tested:

```
function isHTMLInputElement(elem:Element): is HTMLInputElement {
 return elem instanceof HTMLInputElement;
}

// usage
if (isHTMLInputElement(e.target)) { ...do stuff ...}
```

ember-headless-form/src/components/-private/field.hbs
```diff
@@ -5,8 +5,14 @@
       input=(component
         (ensure-safe-component this.InputComponent)
         fieldId=uuid
-        value=this.value
-        setValue=setValue
+        value=this.valueAsString
+        setValue=this.setValue
+      )
+      checkbox=(component
+        (ensure-safe-component this.CheckboxComponent)
```
nicolechung, line 12: Question: what does `ensure-safe-component` do?

ember-headless-form/src/components/-private/control/input.hbs
```diff
@@ -1,6 +1,6 @@
 <input
   type={{@type}}
-  value={{this.valueAsString}}
+  value={{@value}}
   id={{@fieldId}}
   ...attributes
```
nicolechung, line 5: Question: what happens if someone adds their own `{{on 'input' this.handleInput}}` to this element? Or is this a private component that won't be used directly by the consumer?

ember-headless-form/src/components/-private/control/input.ts
```diff
@@ -28,35 +28,37 @@ export type InputType =
...
+        // TS would guard us against using an unsupported `InputType`, but for JS consumers we add a dev-only runtime check here
+        (args.type as string) !== 'checkbox'
     );
 
-    return this.args.value;
+    super(owner, args);
+  }
+
+  get type(): InputType {
+    return this.args.type ?? 'text';
```
nicolechung, line 57: Question, nit: is there a way to set the `args.type` default as `text`? Just wondering if there is a way to hint to the user that `text` is the default type without them having to scroll down to this getter.

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -230,12 +184,73 @@ module('Integration Component headless-form', function (hooks) {
         assert.dom('input').hasAttribute('type', type, `supports type=${type}`);
```
nicolechung, line 184: Should we assert that the type here is `text`?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -230,12 +184,73 @@ module('Integration Component headless-form', function (hooks) {
...
+        );
+      });
+
+      const data = { checked: false };
+
+      await render(<template>
+        <HeadlessForm @data={{data}} as |form|>
+          <form.field @name="checked" as |field|>
+            {{! @glint-expect-error }}
+            <field.input @type="checkbox" />
```
nicolechung, line 204: Nice test!

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -230,12 +184,73 @@ module('Integration Component headless-form', function (hooks) {
...
+          </form.field>
+        </HeadlessForm>
+      </template>);
+    });
+  });
+
+  module('field.checkbox', function () {
+    test('field yields checkbox component', async function (assert) {
+      const data: { checked?: boolean } = {};
```
nicolechung, line 213: Why the `optional` `?` here?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -177,7 +131,7 @@ module('Integration Component headless-form', function (hooks) {
 
   module('field.input', function () {
     test('field yields input component', async function (assert) {
-      const data = { firstName: 'Simon' };
+      const data: { firstName?: string } = {};
```
nicolechung, line 134: Question: why is firstName optional here? On line 253 it is not. Just curious what the difference is.

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -355,25 +379,35 @@ module('Integration Component headless-form', function (hooks) {
...
         await fillIn('input[data-test-first-name]', 'Nicole');
         await fillIn('input[data-test-last-name]', 'Chung');
+        await click('input[data-test-terms]');
         await click('[data-test-submit]');
 
         assert.deepEqual(
           data,
-          { firstName: 'Tony', lastName: 'Ward' },
+          { firstName: 'Tony', lastName: 'Ward', acceptTerms: false },
           'data is not mutated'
```
nicolechung, line 402: Suggestion: `'data is not mutated'` -> 'original data is not mutated'

test-app/tests/integration/components/headless-form-test.gts
nicolechung: oh NVM I see that it's an empty object 🤦

test-app/tests/integration/components/headless-form-test.gts
nicolechung: oh NVM I see that it's an empty object 🤦

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -412,4 +446,51 @@ module('Integration Component headless-form', function (hooks) {
...
+
+  module('Glint', function () {
+    // These tests are not testing any new run-time behaviour that isn't tested elsewhere already.
+    // Rather they are here to make sure they pass glint checks, testing for their types constraints to work as expected
+    // Note: @glint-expect-error behaves just as @ts-expect-error in that in surpresses an error when we *expect* it to error,
+    // but it *also* fails when no expected error is actually present!
+
+    test('@name argument only expects keys of @data', async function (assert) {
+      assert.expect(0);
+      const data = { firstName: 'Simon' };
```
nicolechung, line 458: Suggestion, nit: I'm a bit slow this morning, but it took me a few reads to see that this errors because `lastName` is not supplied to `data`.

Maybe leave some comments here along the lines of `note that `lastName` does not exist here?`.

nicolechung (approved):
This looks great! I left a few small comments/questions/suggestions.

ember-headless-form/src/components/-private/control/input.hbs
NullVoxPopuli: it would layer on top of this `{{on 'input'}}` -- so there would be two event listeners.

We want the element's API to be public, too

ember-headless-form/src/components/-private/control/input.ts
NullVoxPopuli: not yet, but there is a followup proposal to `<template>` that adds exactly this

ember-headless-form/src/components/-private/control/input.ts
simonihmig: Yeah, makes sense. I used `assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);` now though, as Ember will strip that off a production build, and it does the same type assertion as your `is HTMLInputElement` example.

ember-headless-form/src/components/-private/field.hbs
simonihmig: Under Embroider, the use of the `component` helper (which we need here for currying of arguments) will throw a build-time error if you pass a non-string as the first argument, as it cannot statically analyze that to ensure the requested component is pulled into the bundle. But I didn't want to make the yielded component resolvable through a global string (like `headless-form/control/checkbox`), as the component should only be used as a contextual component, and those global string lookups should go away anyway in the future with the newer template-imports.

`ensure-safe-component` is the Embroider-endorsed escape hatch here, which makes Embroider not fail the build, but it will still make sure that what you passed to it is already something that has been imported elsewhere, and as such is already part of the Embroider (strict-mode) bundle. For the detailed explanation of its use cases, see https://github.com/embroider-build/embroider/blob/main/REPLACING-COMPONENT-HELPER.md

ember-headless-form/src/components/-private/control/input.hbs
simonihmig: > Or is this a private component that won't be used directly by the consumer?

We yield the component, so yes, a user could do `<field.input {{on "input" this.doSomething}}/>`. But as Preston said, you just end up with two separate event listeners. Which should be able to peacefully live together...

Unless one of those does something like `event.stopImmediatePropagation()`, but then you are basically begging for problems! 😅

ember-headless-form/src/components/-private/control/input.ts
simonihmig: > but there is a followup proposal to <template> that adds exactly this

Uh, which is this? Have a link at hand?

test-app/tests/integration/components/headless-form-test.gts
simonihmig: We are inside a loop here that asserts that we have a `type` of _all_ supported types, not just `text` (but including it). Or did I misunderstand you?

ember-headless-form/src/components/-private/control/input.ts
NullVoxPopuli: default args to `<template>`, I think the ideal is something like
```js
<template signature={TheSignature} defaultArgs={defaults}>

</template>
```
this _greatly_ reduces boilerplate for template-only components, too.
No link yet, as it's not written. Not sure of @chriskrycho or @dfreeman are going to write it, or maybe it's too early for that kind of planning atm, as they're focused on v1 of Glint atm.

test-app/tests/integration/components/headless-form-test.gts
simonihmig: applied :+1:

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Added :+1:

test-app/tests/integration/components/headless-form-test.gts
nicolechung: Nope my bad seems like I misread the code.

ember-headless-form/src/components/-private/control/input.ts
nicolechung: Thanks for the clarification!

PR #16, opened 2023-01-23
Support radio control

### LINKED ISSUE #7: Support for `<input type="radio">` form control

simonihmig:
Closes #7 

- [x] based on top of #15, needs to be merged first


simonihmig:
Note that this makes the assumption that the values to choose from are strings (given Glint types), as that is what the `value` of `<input>` is, so it aligns with how plain HTML forms work. However (in a future iteration? )we will probably also want to allow choosing between arbitrary objects (e.g. when the form field is used to assign a mode to a `belongsTo` ember-data relationship). In which case we would have to decide what to pass to `<input type="radio" value=<what?>`, simply the object itself and rely on it having a meaningful `toString()` so it does not render as `<input type="radio" value="[object Object]">`? 

I mean the `value` in DOM is not _really_ needed for us, as we have that available in our component JS in unserialized form anyway. But newer frameworks (remix, sveltekit) have been making use of proper form markup to support even form submissions without JS (which we don't have yet in Ember ootb), so I was thinking to make sure we emit "valid" (in that sense) form markup from the beginning. But maybe I am just overcomplicating things here? 😅

NullVoxPopuli:
> we will probably also want to allow choosing between arbitrary objects

we could do this as a serialize / deserialize step at the form level, and not expose it elsewhere? I think it would simplify implementation quite a bit if the `Form` had the logic for "How to get a value for field named X"

> so I was thinking to make sure we emit "valid" (in that sense) form markup from the beginning. But maybe I am just overcomplicating things here? 😅

Not overcomplicating at all, this is an important thing to get right, and most modern form implementations totally forget to make their forms actually valid.
The serialize/deserialize stuff at the `<Form>` level would likely be what someone want want to implement anyway at the server-endpoint / rails-controller level anyway

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -393,19 +558,26 @@ module('Integration Component headless-form', function (hooks) {
...
+            acceptTerms: false,
+          },
           'data is not mutated'
         );
 
         assert.true(
           submitHandler.calledWith({
             firstName: 'Nicole',
             lastName: 'Chung',
+            gender: 'female',
```
nicolechung, line 580: Question: do we need to have an assert explicitly for `FormData` or does the assert on the `submitHandler` effectively check for this?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -185,27 +185,29 @@ module('Integration Component headless-form', function (hooks) {
...
-            {{! @glint-expect-error }}
-            <field.input @type="checkbox" />
-          </form.field>
-        </HeadlessForm>
-      </template>);
-    });
+        await render(<template>
+          <HeadlessForm @data={{data}} as |form|>
+            <form.field @name="checked" as |field|>
+              {{! @glint-expect-error }}
```
NullVoxPopuli, line 204: what's Glint doing here?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -247,10 +249,141 @@ module('Integration Component headless-form', function (hooks) {
     });
   });
 
+  module('field.radio', function () {
```
NullVoxPopuli, line 252: radio group next? :D

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -247,10 +249,141 @@ module('Integration Component headless-form', function (hooks) {
...
+              <radio.input />
+              <radio.label>Foo</radio.label>
+            </field.radio>
+          </form.field>
+        </HeadlessForm>
+      </template>);
+
+      assert.dom('input').hasAttribute(
+        'id',
+        // copied from https://ihateregex.io/expr/uuid/
```
NullVoxPopuli, line 339: nice

ember-headless-form/src/components/-private/control/input.ts
```diff
@@ -47,7 +47,7 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
       `input component does not support @type="${args.type}" as there is a dedicated component for this. Please use the \`field.${args.type}\` instead!`,
       args.type === undefined ||
         // TS would guard us against using an unsupported `InputType`, but for JS consumers we add a dev-only runtime check here
-        (args.type as string) !== 'checkbox'
+        !['checkbox', 'radio'].includes(args.type as string)
```
nicolechung, line 50: Oh so all these `as` casts are JS? I thought `as` was only a TS word

test-app/tests/integration/components/headless-form-test.gts
simonihmig: yeah, well... I have a checkbox group on my (mental) backlog, but what would a radio group be?

ember-headless-form/src/components/-private/control/input.ts
simonihmig: Yeah, exactly.

test-app/tests/integration/components/headless-form-test.gts
NullVoxPopuli: I guess the default way a radio input should be used? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

But! I just found a the test that covers this case: https://github.com/CrowdStrike/ember-headless-form/pull/16/files/dd2315f954edc127e74df0afb01b3ec6b033461f#diff-81f4c9b55c57a592e716604c633b7f95d5a5554742954960cb7bd63be78980ecR398 

So nothing to do! (other than docs, later, I suppose (I'm working on a generator for that part 🙃 ))

ynotdraw (approved):

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Yeah, that's what I meant basically, radios only work as a group, a single radio makes no sense. For checkboxes that's different, we already have a single checkbox (representing a boolean), but a group of checkboxes (representing an array of string values - or a `<Record<string, boolean>`? 🤔 ) is also a use case we should support.

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Sorry, not sure what you mean here?

simonihmig:
I'll go ahead and merge this, as I need this to finalize the other PR, and I think the reviews that have no formal approval here had merely just some questions, more or less related to the PR itself... If you still find something that you feel needs to get addressed, then I am happy to follow-up on this with another PR though!

simonihmig:
_Actually_, after looking into how to combine this (radios) with validation, it seems there is some additional work (investigation + implementation) here to make radios fully accessible. Leaving this out for now, but tracking here: https://github.com/CrowdStrike/ember-headless-form/issues/22

PR #17, opened 2023-01-23
Support textarea control

### LINKED ISSUE #5: Support for `<textarea>` form control

simonihmig:
Closes #5 

- [x] based on top of #15, needs to be merged first

### DIFF
--- a/ember-headless-form/src/components/-private/control/textarea.hbs
+++ b/ember-headless-form/src/components/-private/control/textarea.hbs
@@ -1,0 +1,5 @@
+<textarea
+  id={{@fieldId}}
+  ...attributes
+  {{on 'input' this.handleInput}}
+>{{@value}}</textarea>
--- a/ember-headless-form/src/components/-private/control/textarea.ts
+++ b/ember-headless-form/src/components/-private/control/textarea.ts
@@ -1,0 +1,18 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export interface HeadlessFormControlTextareaComponentSignature {
+  Element: HTMLTextAreaElement;
+  Args: {
+    value: string;
+    fieldId: string;
+    setValue: (value: string) => void;
+  };
+}
+
+export default class HeadlessFormControlTextareaComponent extends Component<HeadlessFormControlTextareaComponentSignature> {
+  @action
+  handleInput(e: Event | InputEvent): void {
+    this.args.setValue((e.target as HTMLTextAreaElement).value);
+  }
+}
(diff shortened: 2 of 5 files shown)


NullVoxPopuli (approved):
Looks good!

ember-headless-form/src/components/-private/control/textarea.ts
```diff
@@ -0,0 +1,18 @@
...
+    value: string;
+    fieldId: string;
+    setValue: (value: string) => void;
+  };
+}
+
+export default class HeadlessFormControlTextareaComponent extends Component<HeadlessFormControlTextareaComponentSignature> {
+  @action
+  handleInput(e: Event | InputEvent): void {
+    this.args.setValue((e.target as HTMLTextAreaElement).value);
```
nicolechung, line 16: Suggestion, nit: again instead of a cast it would be nice to use a custom type guard here...I think one benefit is folks could re-use the custom type guard instead of always reaching for a cast

ember-headless-form/src/components/-private/control/textarea.hbs
```diff
@@ -0,0 +1,5 @@
+<textarea
+  id={{@fieldId}}
+  ...attributes
```
nicolechung, line 3: Question, curious: I see in other component systems there are options to `resize` vertically and horizontally, is this something we'd think of doing at this level or above the `headless` layer?

nicolechung (approved):
Looks good to me!

ember-headless-form/src/components/-private/control/textarea.ts
simonihmig: As in the same case in the other PR, I would add an `assert()` here and remove the type cast, even if that is a bit more verbose than a reusable util function. But `assert()` gets stripped from a production build. And given that the only case where `e.target` is not a `<textarea>` is when _we_ mess this up (by applying `{{on}}` on the wrong element) I would not want to add a runtime overhead by calling a function that is not stripped away...

ember-headless-form/src/components/-private/control/textarea.hbs
simonihmig: Hm, I don't really have prior experience with this, but don't browsers add this ootb now? Like even with a plain `<textarea>` in Chrome there is this dragging "button" on the lower right corner.

But even when there were some more custom solutions to this, I would think that is not something we need to do at this headless level.

NullVoxPopuli: agreed -- resize is automatic -- we can control horiz / vert-only resizing if we want to, but I don't think we need to here.

PR #18, opened 2023-01-23
Fix rollup failing w/ floating deps in CI

simonihmig:
Out of nowhere, CI runs for the floating dependencies scenario [started to fail](https://github.com/CrowdStrike/ember-headless-form/actions/runs/3986007904/jobs/6842377338), for all PRs, even when re-running an older previously green run.

Somehow I couldn't reproduce locally (pinning to same node version, and `pnpm i --no-lockfile`). But I was able to guess the culprit was `rollup-plugin-ts`, as the only rollup related update of dependencies was the fresh new release of [3.2.0](https://github.com/wessberg/rollup-plugin-ts/releases/tag/v3.2.0). 

And indeed...
* switching to node 18 in CI fixes the issue
* they seem to have realised that their version was not working with node 14 anymore (see https://github.com/wessberg/rollup-plugin-ts/compare/v3.2.0...master), which should then have been a breaking change... 🙄

The switch to a newer node version should be unproblematic here, as 
* unlike v1 addons we don't really have a dependency on node for v2 addons, it only plays a role in our own prepublish build step, but not for consumers
* we haven't released this anyway 😅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 18.x
           cache: pnpm
       - name: Install Dependencies
         run: pnpm install --frozen-lockfile
@@ -46,7 +46,7 @@ jobs:
           version: 7
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 18.x
           cache: pnpm
       - name: Install Dependencies
         run: pnpm install --no-lockfile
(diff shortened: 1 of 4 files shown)


simonihmig:
FWIW, this will nevertheless cause a bit of churn in the v2 addon space using this plugin. See e.g. https://github.com/simonihmig/ember-stargate/pull/614

simonihmig:
@ynotdraw this might also affect your work maybe?

NullVoxPopuli (approved):

NullVoxPopuli:
All my repos use [this](https://github.com/NullVoxPopuli/ember-resources/blob/main/.github/actions/pnpm/action.yml#L9), which supports volta, so I think as long as v2 addons are using volta, they should be safe from the rollup-plugin-ts changes

ynotdraw:
> @ynotdraw this might also affect your work maybe?

Thank you! 🙇

Issue #9, opened 2023-01-13
Support for validation API


simonihmig:
See https://gist.github.com/simonihmig/ff39facb915d41f4b27717a10e2196e9#custom-validations

Issue #13, opened 2023-01-13
Support for async submit action and yielding async state


simonihmig:
See https://gist.github.com/simonihmig/ff39facb915d41f4b27717a10e2196e9#async-support

simonihmig:
In addition to async submit handlers, we could also integrate support for async validation (see https://github.com/CrowdStrike/ember-headless-form/pull/21#discussion_r1087038441), by yielding the async state, so users can show the pending validation somehow.

PR #21, opened 2023-01-25
Add validation API

simonihmig:
This is the first part of #9. It adds support for the basic validation API (as outlined in the [initial API draft](https://gist.github.com/simonihmig/ff39facb915d41f4b27717a10e2196e9)), allowing for validation callbacks to be registered using `@validate`, both on the form and the field level. Although the `@validateOn` and `@revalidateOn` arguments have been added here already, they are not supported yet. So the form only validates on submit for now, as a regular server-side rendered form would do. Support for those variations will come in another PR, as this one seems already fairly large enough.

This also incorporates some of the insights we gained from yesterday's a11y workshop with Noah:
* apply `aria-invalid` to controls
* associate controls to error messages using `aria-errormessage` (and `aria-live`)

To Do:

- [x] wait for #16 to get merged
- [x] ~wire-up radio controls w/ a11y attributes as well~ will follow in #22


ember-headless-form/src/components/-private/control/textarea.hbs
```diff
@@ -1,5 +1,7 @@
 <textarea
   id={{@fieldId}}
+  aria-invalid={{if @invalid 'true'}}
```
nicolechung, line 3: 💯 !!

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -494,6 +514,837 @@ module('Integration Component headless-form', function (hooks) {
...
+          type: 'required',
+          value,
+          message: `${field} is required!`,
+        });
+      } else {
+        if (value.charAt(0).toUpperCase() !== value.charAt(0)) {
+          errors.push({
+            type: 'uppercase',
+            value,
+            message: `${field} must be upper case!`,
```
nicolechung, line 574: Question, total nit: if the first letter starts with a capital, isn't that something like sentence case and not uppercase?

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -55,6 +55,26 @@ module('Integration Component headless-form', function (hooks) {
...
     });
 
+    test('@name must be unique', async function (assert) {
+      assert.expect(1);
+      const data = { firstName: 'Simon' };
+
+      setupOnerror((e: Error) => {
+        assert.strictEqual(
+          e.message,
+          'Assertion Failed: You passed @name="firstName" to the form field, but this is already in use. Names of form fields must be unique!',
```
NullVoxPopuli, line 65: ooo this is exciting!!!

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -494,6 +514,837 @@ module('Integration Component headless-form', function (hooks) {
...
+    };
+
+    const validateFieldCallbackAsync = async (
+      value: string | undefined,
+      field: string
+    ) => {
+      return await validateFieldCallbackSync(value, field);
+    };
+
+    module('form @validation callback', function () {
```
NullVoxPopuli, line 597: it may be worth it to split this into a separate test file

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -494,6 +514,837 @@ module('Integration Component headless-form', function (hooks) {
...
+              .hasText('Foo is an invalid first name!');
+
+            assert.dom('[data-test-last-name-errors]').doesNotExist();
+          });
+        })
+      );
+    });
+  });
+
   module('Glint', function () {
```
NullVoxPopuli, line 1348: might be worth it to extract this to a separate file -- these files are getting quite long, and when it comes time to do any sort of automated migration on these files, having them be a smidge smaller allows for better incremental migration.

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -494,6 +514,837 @@ module('Integration Component headless-form', function (hooks) {
...
+      value: string | undefined,
+      field: string
+    ) => {
+      return await validateFieldCallbackSync(value, field);
+    };
+
+    module('form @validation callback', function () {
+      [
+        { testType: 'sync', validateCallback: validateFormCallbackSync },
+        { testType: 'async', validateCallback: validateFormCallbackAsync },
```
NullVoxPopuli, line 600: for tests using the async callback, how would someone know to show pending state in the UI while they wait for the validation to complete?

NullVoxPopuli (commented):
Looking good -- left a few questions.

Also, would like to see tests using the native APIs.

https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#the_required_attribute

nicolechung (approved):
💯

test-app/tests/integration/components/headless-form-test.gts
```diff
@@ -494,6 +514,837 @@ module('Integration Component headless-form', function (hooks) {
...
+
+            await render(<template>
+              <HeadlessForm
+                @data={{data}}
+                @onSubmit={{submitHandler}}
+                as |form|
+              >
+                <form.field
+                  @name="firstName"
+                  @validate={{validateCallback}}
```
ynotdraw, line 1103: 😱 perfect! 💯

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Yeah, probably. As a non-native speaker, I was not aware that "sentence case" was a phrase for that. 😅

Given this is only for tests, I would tend to not change this, as it would require changes in a lot of places, unless you insist? :wink:

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Yeah, I wondered about this too, and had no clear idea how to structure this best.

So, when following the existing Ember convention of having `component/foo-test.js` for component `foo`, then this is what you end up, as `<HeadlessForm>` is the only public component the addon exposes. In other projects (e.g. ember-bootstrap) I had tests for the yielded components in separate files, like here it could be `components/headless-form/field.gts`, but neither directly accessing the component as `<HeadlessForm::Field>` (which wouldn't work currently, and be pretty terrible actually as that is not testing the real API) no putting tests that use `<HeadlessForm as |form|>` into a test file that is specific for the sub-components felt right.

But yeah, I think we don't have to follow the convention strictly here, and splitting this up into separate features of the same component as you suggested (like basics, validation, glint) would make sense. Will do that!

test-app/tests/integration/components/headless-form-test.gts
simonihmig: Good point! 

There is https://github.com/CrowdStrike/ember-headless-form/issues/13, which was mainly about supporting asynchrony of the submit action itself, but I guess we could/should also account for async validation as part of that as well! Will add a note!

simonihmig:
> Also, would like to see tests using the native APIs.

Yeah, this is not supported _yet_, but will come soon(TM). I track that in #10

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -0,0 +1,833 @@
...
+              '[data-test-first-name-errors] [data-test-error="1"] [data-test-error-message]'
+            )
+            .hasText('Foo is an invalid first name!');
+
+          assert.dom('[data-test-last-name-errors]').doesNotExist();
+        });
+      })
+    );
+  });
+});
```
ynotdraw, line 833: Maybe this is a bit redundant or not necessary, but I'm curious what you think.

Would it be helpful to include a test that _does not_ use the field components directly, but provides their own input styled / opinionated components?  This is mostly to cover my exact use case, where we will use controls from a different addon rather than style the controls provided by headless-form.

```js
await render(
  <template>
    <HeadlessForm @data={{data}} as |form|>
      <form.field
        @name="firstName"
        @validate={{validateCallback}}
        as |field|
      >
        <!-- Verify validation works as expected for this -->
        <CustomInputComponent
          @label="First Name"
          id={{field.id}}
          value={{field.value}}
        />
        <field.errors data-test-first-name-errors />
      </form.field>
      <button type="submit" data-test-submit>Submit</button>
    </HeadlessForm>
  </template>
);
```

```hbs
<!-- CustomInputComponent -->
<!-- not complete example -->
<div>
  <label>{{@label}}</label>
  <input ...attributes />
</div>
```

Having an example of using custom control elements may help when it comes adoption time as well (i.e., have a test/docs to look at how to do this exact case).

NullVoxPopuli (approved):
There are some follow-up things mentioned, but I'm happy with the state of this, given the desire to incrementally improve!
 - https://github.com/CrowdStrike/ember-headless-form/issues/10 - native support
 - test file splitting, for being less overwhelmed: https://github.com/CrowdStrike/ember-headless-form/pull/21#discussion_r1087706802
 - and then @ynotdraw 's suggestion, https://github.com/CrowdStrike/ember-headless-form/pull/21#discussion_r1087916485, which plays in to #10, I think

test-app/tests/integration/components/headless-form-validation-test.gts
nicolechung: I too would like this...but maybe this can be a follow up since this PR is getting quite large already.

simonihmig: Yeah, we should definitely add something like that into our future docs. For the test here, the input (whether that's built-in or custom) does not really play a role with validation, so the test wouldn't have much value here beyond documentation purposes I think.

However I am just realizing that using `field.errors` is actually required to use, otherwise you won't get access to the validation errors. Previously `field.errors` was just an array of errors, but now (in one of the later changes) I refactored this yield a component that actually renders a div, and that one yields the raw errors (see [example](https://github.com/CrowdStrike/ember-headless-form/pull/21/files#diff-aa61a3f8763adc87f435e3907b140da6f0c4f85fb8b1747114c4163d4bb64a0aR355-R380)). The reason was yesterday's a11y session basically, as this allows us to bake some of the a11y best practices into that div (using `aria-errormessage` to refer to is), see [this test](https://github.com/CrowdStrike/ember-headless-form/pull/21/files#diff-aa61a3f8763adc87f435e3907b140da6f0c4f85fb8b1747114c4163d4bb64a0aR270-R321). So if this were truly headless, then you wouldn't have anything to add the a11y markup to.

So given that, I wonder if there might be a need to not even use that, but render 100% of the markup yourself. In which case we would need to expose the raw errors directly again, maybe as `field.validationErrors`? Or rename what we have now from `field.errors` to something else? :think:

I think we have to block this PR though, we will still have a good amount of iterations before even thinking about using this and cutting a release...

NullVoxPopuli: as long as https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/validity is supported, in the future, we're 💯

ynotdraw (approved):
This is coming along nicely!

PR #23, opened 2023-01-27
Fix rendering `name` attribute, reorganize internal types

simonihmig:
Previous work missed to properly set `name` on each form control (inputs) for proper form markup. This was not needed for functionality, but we should strive for proper markup.

Unfortunately this triggered also a cascade of type refactorings, as for internal Glint types to all pass, `name` (which maps to a key of the passed `@data`) must be a string, and not any other type that `keyof` can be (number or symbol). Now we are a bit stricter about types (and more verbose 🙈), dealing only with string-typed keys and such subsets of form data.

### DIFF
--- a/ember-headless-form/src/components/-private/types.ts
+++ b/ember-headless-form/src/components/-private/types.ts
@@ -1,0 +1,84 @@
+/**
+ * What the user can pass as @data
+ */
+export type UserData = object;
+
+/**
+ * The subset of properties of DATA, whose keys are strings (and not number or symbol)
+ * Only this data is useable in the form
+ */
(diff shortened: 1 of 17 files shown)


NullVoxPopuli:
`name` is required for FormData use cases, so this is a great change!

ember-headless-form/src/components/-private/types.ts
```diff
@@ -0,0 +1,84 @@
+/**
```
NullVoxPopuli, line 1: thanks for adding JSDoc to all these! 

JSDoc is soooooo important, cause it shows up in everyone's editors, and super aids in helping folks understand what they're looking at

NullVoxPopuli (approved):
💯

PR #24, opened 2023-01-27
Support `@validateOn="blur/change"`

simonihmig:
This is a draft PR to just start a discussion, will close it afterwards as it is _not_ ready for review yet...

Let's say we have `@validateOn="blur"`. Now on an input (or other field control) the user causes the `blur`/`focusout` events to trigger. The action that must follow is that we validate, all fields or just this one, but most importantly we only _show_ the field's errors that triggered the event, not any other. 

So the question is this: how do we know, which field caused the event? There are several ways to do this, all with their pro and cons:

1. we can let the event bubble up to `<form>` (so our main `<HeadlessForm>` component), inspect the event target's `name` attribute (which is what we passed as `form.field @name="..."`) and use that to look-up the field. This. is what this PR currently does, and it works so far. However it requires the event to be triggered from a real native form control, and it needs to have `name` properly set up. That's the case here, but this might not work everywhere, e.g. might not work for rich custom controls like a `<PowerSelect>`, or combined controls like a date picker that (internally) is split up into 3 `<select>` with `name="date_{day,month,year}"`, so the `name` won't match what the field's name is (just `date` in this example). I think it would be possible to provide a yielded action like `triggerValidationFor(name)`, but this requires actively wiring this up for the user.
2. we could _only_ do the latter, i.e. provide a `triggerValidationFor(name)` and all controls need to wire this up explicitly, so our own here (`field.input`, `field.checkbox` etc.) same as any custom ones., and not rely on any `name` matching logic at all
3. or we could make our `form.field` component create a wrapper div, and attach the event listeners to that. Currently it does not render any markup (truly headless), so when the event bubbles up from the `<input>` it would immediately reach the `<form>`, at which point we wouldn't know which field this belongs to (unless we do 1.). But with a wrapping div (which would have `...attributes`, so users could use it for their own styling etc.) we would be able to catch the event within the boundaries of the field, so we _know_ which field this event belongs to, and show validation only for this field. This has the downside of an additional wrapping div, that does not have any other good reason to exist like a11y or so, so not really headless in a pure way. But the upside is that this would work out of the box, both for our own controls, but also any custom ones, as long as they are able to trigger `focusout`/`change` events properly.

Any thoughts?

### DIFF
--- a/ember-headless-form/src/components/headless-form.ts
+++ b/ember-headless-form/src/components/headless-form.ts
@@ -1,6 +1,7 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
+import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 
 import { TrackedObject } from 'tracked-built-ins';
@@ -10,11 +11,13 @@ import FieldComponent from './-private/field';
 import type { HeadlessFormFieldComponentSignature } from './-private/field';
 import type {
   ErrorRecord,
-  FieldData,
+  FieldRegistrationData,
+  FieldValidateCallback,
   FormData,
   FormKey,
   FormValidateCallback,
   UserData,
+  ValidationError,
 } from './-private/types';
 import type { ComponentLike, WithBoundArgs } from '@glint/template';
 
@@ -37,12 +40,25 @@ export interface HeadlessFormComponentSignature<DATA extends UserData> {
     default: [
       {
         field: WithBoundArgs<
-          typeof FieldComponent<FormData<DATA>>,
+          typeof FieldComponent<DATA>,
           'data' | 'set' | 'errors' | 'registerField' | 'unregisterField'
         >;
       }
     ];
   };
+}
+
+class FieldData<
+  DATA extends FormData,
+  KEY extends FormKey<DATA> = FormKey<DATA>
+> {
+  constructor(fieldRegistration: FieldRegistrationData<DATA, KEY>) {
+    this.validate = fieldRegistration.validate;
+  }
+
+  @tracked validationEnabled = false;
+
+  validate?: FieldValidateCallback<DATA, KEY>;
 }
 
 export default class HeadlessFormComponent<
@@ -51,13 +67,15 @@ export default class HeadlessFormComponent<
   FieldComponent: ComponentLike<HeadlessFormFieldComponentSignature<DATA>> =
     FieldComponent;
 
-  internalData: FormData<DATA> = new TrackedObject(
-    this.args.data ?? {}
-  ) as FormData<DATA>;
+  // we cannot use (modifier "on") directly in the template due to https://github.com/emberjs/ember.js/issues/19869
+  on = on;
+
+  internalData: DATA = new TrackedObject(this.args.data ?? {}) as DATA;
 
   fields = new Map<FormKey<FormData<DATA>>, FieldData<FormData<DATA>>>();
 
   @tracked lastValidationResult?: ErrorRecord<FormData<DATA>>;
+  @tracked showAllValidations = false;
 
   get validateOn(): ValidateOn {
     return this.args.validateOn ?? 'submit';
@@ -65,6 +83,16 @@ export default class HeadlessFormComponent<
 
   get revalidateOn(): ValidateOn {
     return this.args.revalidateOn ?? 'change';
+  }
+
+  get fieldValidationEvent(): 'focusout' | 'change' | undefined {
+    const { validateOn } = this;
+
+    return validateOn === 'submit'
+      ? undefined
+      : validateOn === 'blur'
+      ? 'focusout'
+      : validateOn;
   }
 
   get hasValidationErrors(): boolean {
@@ -106,11 +134,38 @@ export default class HeadlessFormComponent<
     return Object.keys(errors).length > 0 ? errors : undefined;
   }
 
+  get visibleErrors(): ErrorRecord<FormData<DATA>> | undefined {
+    if (!this.lastValidationResult) {
+      return undefined;
+    }
+
+    const visibleErrors: ErrorRecord<FormData<DATA>> = {};
+
+    for (const [field, errors] of Object.entries(this.lastValidationResult) as [
+      FormKey<FormData<DATA>>,
+      ValidationError<FormData<DATA>[FormKey<FormData<DATA>>]>[]
+    ][]) {
+      if (this.showErrorsFor(field)) {
+        visibleErrors[field] = errors;
+      }
+    }
+
+    return visibleErrors;
+  }
+
+  showErrorsFor(field: FormKey<FormData<DATA>>): boolean {
+    return (
+      this.showAllValidations ||
+      (this.fields.get(field)?.validationEnabled ?? false)
+    );
+  }
+
   @action
   async onSubmit(e: Event): Promise<void> {
     e.preventDefault();
 
     this.lastValidationResult = await this.validate();
+    this.showAllValidations = true;
 
     if (!this.hasValidationErrors) {
       this.args.onSubmit?.(this.internalData);
@@ -126,7 +181,7 @@ export default class HeadlessFormComponent<
   @action
   registerField(
     name: FormKey<FormData<DATA>>,
-    field: FieldData<FormData<DATA>>
+    field: FieldRegistrationData<FormData<DATA>>
   ): void {
     assert(
       `You passed @name="${String(
@@ -134,7 +189,7 @@ export default class HeadlessFormComponent<
       )}" to the form field, but this is already in use. Names of form fields must be unique!`,
       !this.fields.has(name)
     );
-    this.fields.set(name, field);
+    this.fields.set(name, new FieldData(field));
   }
 
   @action
@@ -145,5 +200,22 @@ export default class HeadlessFormComponent<
   @action
   set<KEY extends FormKey<FormData<DATA>>>(key: KEY, value: DATA[KEY]): void {
     this.internalData[key] = value;
+  }
+
+  @action
+  async handleFieldValidation(e: Event): Promise<void> {
+    const { target } = e;
+    const { name } = target as HTMLInputElement;
+
+    if (name) {
+      const field = this.fields.get(name as FormKey<FormData<DATA>>);
+
+      if (field) {
+        this.lastValidationResult = await this.validate();
+        field.validationEnabled = true;
+      }
+    } else {
+      // @todo how to handle custom controls that don't emit focusout/change events from native form controls?
+    }
   }
 }
--- a/ember-headless-form/package.json
+++ b/ember-headless-form/package.json
@@ -57,6 +57,7 @@
     "@types/ember__debug": "^4.0.0",
     "@types/ember__engine": "^4.0.0",
     "@types/ember__error": "^4.0.0",
+    "@types/ember__modifier": "^4.0.3",
     "@types/ember__object": "^4.0.0",
     "@types/ember__polyfills": "^4.0.0",
     "@types/ember__routing": "^4.0.0",
--- a/ember-headless-form/src/components/-private/types.ts
+++ b/ember-headless-form/src/components/-private/types.ts
@@ -55,7 +55,7 @@ export type FieldValidateCallback<
  * Internal structure to track used fields
  * @private
  */
-export interface FieldData<
+export interface FieldRegistrationData<
   DATA extends FormData,
   KEY extends FormKey<DATA> = FormKey<DATA>
 > {
@@ -69,7 +69,7 @@ export interface FieldData<
 export type RegisterFieldCallback<
   DATA extends FormData,
   KEY extends FormKey<DATA> = FormKey<DATA>
-> = (name: KEY, field: FieldData<DATA, KEY>) => void;
+> = (name: KEY, field: FieldRegistrationData<DATA, KEY>) => void;
 
 export type UnregisterFieldCallback<
   DATA extends FormData,
@@ -79,6 +79,4 @@ export type UnregisterFieldCallback<
 /**
  * Mapper type to construct subset of objects, whose keys are only strings (and not number or symbol)
  */
-export type OnlyStringKeys<T extends object> = {
-  [P in Extract<keyof T, string>]: T[P];
-};
+export type OnlyStringKeys<T extends object> = Pick<T, keyof T & string>;
--- a/ember-headless-form/src/components/headless-form.hbs
+++ b/ember-headless-form/src/components/headless-form.hbs
@@ -1,11 +1,18 @@
-<form ...attributes {{on 'submit' this.onSubmit}}>
+{{! ignoring prettier here is need to *not* wrap the modifier usage below into a new line, making @glint-expect-error fail to work 🙈 }}
+{{! prettier-ignore }}
+<form
+  ...attributes
+  {{on 'submit' this.onSubmit}}
+  {{! @glint-expect-error: modifier helper not supported, see https://github.com/typed-ember/glint/issues/410 }}
+  {{(if this.fieldValidationEvent (modifier this.on this.fieldValidationEvent this.handleFieldValidation))}}
+>
   {{yield
     (hash
       field=(component
         (ensure-safe-component this.FieldComponent)
         data=this.internalData
         set=this.set
-        errors=this.lastValidationResult
+        errors=this.visibleErrors
         registerField=this.registerField
         unregisterField=this.unregisterField
       )
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -39,6 +39,7 @@ importers:
       '@types/ember__debug': ^4.0.0
       '@types/ember__engine': ^4.0.0
       '@types/ember__error': ^4.0.0
+      '@types/ember__modifier': ^4.0.3
       '@types/ember__object': ^4.0.0
       '@types/ember__polyfills': ^4.0.0
       '@types/ember__routing': ^4.0.0
@@ -90,6 +91,7 @@ importers:
       '@types/ember__debug': 4.0.3_@babel+core@7.20.12
       '@types/ember__engine': 4.0.4_@babel+core@7.20.12
       '@types/ember__error': 4.0.2
+      '@types/ember__modifier': 4.0.3_@babel+core@7.20.12
       '@types/ember__object': 4.0.5_@babel+core@7.20.12
       '@types/ember__polyfills': 4.0.1
       '@types/ember__routing': 4.0.12_@babel+core@7.20.12
@@ -2975,10 +2977,7 @@ packages:
   /@types/ember__controller/4.0.4:
     resolution: {integrity: sha512-+f0knTIJJkRX5xijeSI/n4FvLfhMFFxIxODyFFFFB483EryYuts3QzpTwU5D66WQ5rAbZvpPRXRMPTTCNJoUhg==}
     dependencies:
-      '@types/ember__object': 4.0.5
-    transitivePeerDependencies:
-      - '@babel/core'
-      - supports-color
+      '@types/ember__object': 4.0.5_@babel+core@7.20.12
     dev: true
 
   /@types/ember__controller/4.0.4_@babel+core@7.20.12:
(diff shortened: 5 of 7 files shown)


ember-headless-form/src/components/headless-form.ts
```diff
@@ -146,4 +201,21 @@ export default class HeadlessFormComponent<
...
     this.internalData[key] = value;
   }
+
+  @action
+  async handleFieldValidation(e: Event): Promise<void> {
+    const { target } = e;
+    const { name } = target as HTMLInputElement;
+
+    if (name) {
+      const field = this.fields.get(name as FormKey<FormData<DATA>>);
```
simonihmig, line 211: Here is the name matching logic of variant 1 happening!

ember-headless-form/src/components/headless-form.ts
```diff
@@ -146,4 +201,21 @@ export default class HeadlessFormComponent<
...
+
+      if (field) {
+        this.lastValidationResult = await this.validate();
+        field.validationEnabled = true;
+      }
+    } else {
+      // @todo how to handle custom controls that don't emit focusout/change events from native form controls?
```
simonihmig, line 218: And yeah, that's the downside of that approach 😬

NullVoxPopuli:
I think allowing bubbling is one of our requirements, because users may want to do this approach to get [easy / no-fuss two-way binding™](https://limber.glimdown.com/edit?format=glimdown&t=%23%20Forms%0A%0A%23%23%23%20Using%20_The%20Platform_%0A%0AVanilla%20JavaScript%20has%20everything%20we%20need%20to%20handle%20form%20data%2C%20de-sync%20it%20from%20our%20source%20data%20and%20collect%20all%20user%20input%20upon%20submission.%0A%0AOther%20abstractions%2C%20such%20as%20the%20%22changeset%22%20concept%20contain%20a%20lot%20of%20this%20functionality%20and%20have%20additional%20utilities%20such%20as%20rollback%2C%20snapshots%2C%20forking%2C%20etc%2C%20but%20that%20is%20a%20topic%20for%20another%20demo.%0A%0AIn%20the%20form%20below%2C%20we%20create%20a%20Vanilla%E2%84%A2%20%5BHTML%20form%5D%5B2%5D%2C%20and%20only%20add%20%22Ember%22%20code%20for%20handling%20the%20form%20submission%20and%20field%20inputs.%20By%20default%2C%20form%20submissions%20will%20cause%20a%20page%20reload%2C%20so%20in%20a%20single-page-app%2C%20we%20need%20to%20prevent%20that%20default%20behavior.%0A%0AUsing%20the%20native%20API%2C%20%5BFormData%5D%5B1%5D%2C%20we%20can%20gather%20the%20user%20inputs%20when%20the%20user%20presses%20the%20submit%20button.%0A%0A%60%60%60gjs%20live%0Aimport%20%7B%20on%20%7D%20from%20%27%40ember%2Fmodifier%27%3B%0Aimport%20%7B%20cell%20%7D%20from%20%27ember-resources%27%3B%0A%0Aconst%20state%20%3D%20cell()%3B%0A%0Aconst%20handleInput%20%3D%20(event)%20%3D%3E%20%7B%0A%20%20let%20formData%20%3D%20new%20FormData(event.currentTarget)%3B%0A%20%20let%20data%20%3D%20Object.fromEntries(formData.entries())%3B%0A%0A%20%20state.current%20%3D%20JSON.stringify(data%2C%20null%2C%202)%3B%0A%7D%3B%0A%0Aconst%20handleSubmit%20%3D%20(%20event)%20%3D%3E%20%7B%0A%20%20event.preventDefault()%3B%0A%20%20handleInput(event)%3B%0A%7D%3B%0A%0A%3Ctemplate%3E%0A%20%20%3Cform%0A%20%20%20%20%7B%7Bon%20%27input%27%20handleInput%7D%7D%0A%20%20%20%20%7B%7Bon%20%27submit%27%20handleSubmit%7D%7D%0A%20%20%20%20class%3D%22grid%20gap-2%22%0A%20%20%20%20style%3D%22max-width%3A%20300px%22%0A%20%20%3E%0A%20%20%20%20%3Clabel%3E%20First%20Name%0A%20%20%20%20%20%20%3Cinput%20name%3D%27firstName%27%3E%0A%20%20%20%20%3C%2Flabel%3E%0A%0A%20%20%20%20%3Clabel%3E%20Favorite%20Date%0A%20%20%20%20%20%20%3Cinput%20type%3D%27date%27%20name%3D%27favoriteDate%27%3E%0A%20%20%20%20%3C%2Flabel%3E%0A%0A%20%20%20%20%3Cbutton%20type%3D%27submit%27%3ESubmit%3C%2Fbutton%3E%0A%20%20%3C%2Fform%3E%0A%0A%20%20%3Cbr%3E%3Cbr%3E%0A%0A%20%20FormData%3A%0A%20%20%3Cpre%3E%7B%7Bstate.current%7D%7D%3C%2Fpre%3E%0A%0A%20%20%3Cstyle%3E%0A%20%20%20%20input%20%7B%20border%3A%201px%20solid%3B%20%7D%0A%20%20%3C%2Fstyle%3E%0A%3C%2Ftemplate%3E%0A%60%60%60%0A%0A%3Chr%3E%0A%0A%23%23%23%20%22Platform%22%20References%0A%20-%20%5B%60%3Cform%3E%60%20on%20MDN%5D%5B2%5D%0A%20-%20%5B%60FormData%60%20on%20MDN%5D%5B1%5D%0A%0A%23%23%23%20Ember%20References%0A%20-%20%5Bthe%20%60on%60%20modifier%5D%5B3%5D%0A%0A%0A%5B1%5D%3A%20https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FAPI%2FFormData%0A%5B2%5D%3A%20https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FHTML%2FElement%2Fform%0A%5B3%5D%3A%20https%3A%2F%2Fguides.emberjs.com%2Frelease%2Fcomponents%2Fcomponent-state-and-actions%2F%23toc_html-modifiers-and-actions%0A)

ember-headless-form/src/components/headless-form.ts
NullVoxPopuli: imo,
custom controls can / should emit these events tho. 

"To be a custom control, it must behave like a control, including having <these events>"

ynotdraw: > imo,
> custom controls can / should emit these events tho.

Yeah, I agree.  We definitely will want validation to trigger for controls like `<PowerSelect`, `<PowerCalendar`, etc. used inside of forms.  Getting these elements to act more like their native versions would be 💯

simonihmig: Ok, but the question is still whether to do 1. or 3. then.

For 1. to work, the event must have `event.target.name === @name passed to field`. A `<PowerSelect>` could have `<input type="hidden" name={{@name}}>` (does it, idk?), and if that triggers `focusout` then we are fine. 

But what about the example with the 3-select date picker. It will receive `@name="date"`, but will split this up internally to 3 selects of names `date_day`, `date_month`, `date_year`. When one of those selects triggers a `focusout`, our form will not know what field this belongs to, because `event.target.name === 'date'`. 

For 3., `event.target.name` would not matter, and everything would "just work"(TM). But the downside (wrapper div) remains...

NullVoxPopuli: I think the caveat for 1 is fine.
For 3, can you write out the details of this approach by extending [this demo](https://limber.glimdown.com/edit?format=glimdown&t=%23%20Forms%0A%0A%23%23%23%20Using%20_The%20Platform_%0A%0AVanilla%20JavaScript%20has%20everything%20we%20need%20to%20handle%20form%20data%2C%20de-sync%20it%20from%20our%20source%20data%20and%20collect%20all%20user%20input%20upon%20submission.%0A%0AOther%20abstractions%2C%20such%20as%20the%20%22changeset%22%20concept%20contain%20a%20lot%20of%20this%20functionality%20and%20have%20additional%20utilities%20such%20as%20rollback%2C%20snapshots%2C%20forking%2C%20etc%2C%20but%20that%20is%20a%20topic%20for%20another%20demo.%0A%0AIn%20the%20form%20below%2C%20we%20create%20a%20Vanilla%E2%84%A2%20%5BHTML%20form%5D%5B2%5D%2C%20and%20only%20add%20%22Ember%22%20code%20for%20handling%20the%20form%20submission%20and%20field%20inputs.%20By%20default%2C%20form%20submissions%20will%20cause%20a%20page%20reload%2C%20so%20in%20a%20single-page-app%2C%20we%20need%20to%20prevent%20that%20default%20behavior.%0A%0AUsing%20the%20native%20API%2C%20%5BFormData%5D%5B1%5D%2C%20we%20can%20gather%20the%20user%20inputs%20when%20the%20user%20presses%20the%20submit%20button.%0A%0A%60%60%60gjs%20live%0Aimport%20%7B%20on%20%7D%20from%20%27%40ember%2Fmodifier%27%3B%0Aimport%20%7B%20cell%20%7D%20from%20%27ember-resources%27%3B%0A%0Aconst%20state%20%3D%20cell()%3B%0A%0Aconst%20handleInput%20%3D%20(event)%20%3D%3E%20%7B%0A%20%20let%20formData%20%3D%20new%20FormData(event.currentTarget)%3B%0A%20%20let%20data%20%3D%20Object.fromEntries(formData.entries())%3B%0A%0A%20%20state.current%20%3D%20JSON.stringify(data%2C%20null%2C%202)%3B%0A%7D%3B%0A%0Aconst%20handleSubmit%20%3D%20(%20event)%20%3D%3E%20%7B%0A%20%20event.preventDefault()%3B%0A%20%20handleInput(event)%3B%0A%7D%3B%0A%0A%3Ctemplate%3E%0A%20%20%3Cform%0A%20%20%20%20%7B%7Bon%20%27input%27%20handleInput%7D%7D%0A%20%20%20%20%7B%7Bon%20%27submit%27%20handleSubmit%7D%7D%0A%20%20%20%20class%3D%22grid%20gap-2%22%0A%20%20%20%20style%3D%22max-width%3A%20300px%22%0A%20%20%3E%0A%20%20%20%20%3Clabel%3E%20First%20Name%0A%20%20%20%20%20%20%3Cinput%20name%3D%27firstName%27%3E%0A%20%20%20%20%3C%2Flabel%3E%0A%0A%20%20%20%20%3Clabel%3E%20Favorite%20Date%0A%20%20%20%20%20%20%3Cinput%20type%3D%27date%27%20name%3D%27favoriteDate%27%3E%0A%20%20%20%20%3C%2Flabel%3E%0A%0A%20%20%20%20%3Cbutton%20type%3D%27submit%27%3ESubmit%3C%2Fbutton%3E%0A%20%20%3C%2Fform%3E%0A%0A%20%20%3Cbr%3E%3Cbr%3E%0A%0A%20%20FormData%3A%0A%20%20%3Cpre%3E%7B%7Bstate.current%7D%7D%3C%2Fpre%3E%0A%0A%20%20%3Cstyle%3E%0A%20%20%20%20input%20%7B%20border%3A%201px%20solid%3B%20%7D%0A%20%20%3C%2Fstyle%3E%0A%3C%2Ftemplate%3E%0A%60%60%60%0A%0A%3Chr%3E%0A%0A%23%23%23%20%22Platform%22%20References%0A%20-%20%5B%60%3Cform%3E%60%20on%20MDN%5D%5B2%5D%0A%20-%20%5B%60FormData%60%20on%20MDN%5D%5B1%5D%0A%0A%23%23%23%20Ember%20References%0A%20-%20%5Bthe%20%60on%60%20modifier%5D%5B3%5D%0A%0A%0A%5B1%5D%3A%20https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FAPI%2FFormData%0A%5B2%5D%3A%20https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FHTML%2FElement%2Fform%0A%5B3%5D%3A%20https%3A%2F%2Fguides.emberjs.com%2Frelease%2Fcomponents%2Fcomponent-state-and-actions%2F%23toc_html-modifiers-and-actions%0A) -- i just want to be _super extra sure_ I understand how this wrapping div approach would work

ynotdraw:
For number 3 above -   I could be misunderstanding this, but thought I'd ask.  One concern I'd have here is if the markup is:

```hbs
<div ...attributes>
  <input />
</div>
```

rather than

```hbs
<div>
  <input ...attributes />
</div>
```

It's my understanding that you have no way to access the input component directly to add modifiers or attributes without exposing component arguments, right?  So if I wanted to do `<input type="number"`, there'd have to be a component argument to pass that into the input element directly, right?

```hbs
<div ...attributes>
  <input type={{@type}} />
</div>
```

As a user, I'd much rather have access directly to the input element with the attributes spread so that I can provide attributes and modifiers that just work.

```hbs
<SimonsInput type="number" {{on "focus" this.someAction}} />
```

So I think for me, I'm leaning more towards option 1 knowing that, but also am curious what @NullVoxPopuli and @nicolechung think.

ember-headless-form/src/components/headless-form.ts
simonihmig: Not sure tbh how the demo would relate to this addon, but I'll try it this way:

The current [form.field template](https://github.com/CrowdStrike/ember-headless-form/blob/main/ember-headless-form/src/components/-private/field.hbs) would have a wrapping div like so:

```hbs
<div {{on "focusout (fn @triggerValidationFor @name)}} ...attributes>
  {{! here is what is currently in the template }}
</div>
```

`@triggerValidationFor` would be curried into the component when yielding it as a contextual component, so that's a private implementation detail.

From the caller's site:

```hbs
<HeadlessForm as |form|>
  <form.field @name="date" as |field|>  <-- this would now be a div that captures the event, and *knows* what @name it belongs to!
    <field.label>Date:</field.label>
    <CustomDatePickerComponentConsistingOfThreeSelectsWithUnknownNames
      @value={{field.value}}
      @onUpdate={{field.setValue}}
    />
    <field.errors/>
  </form
</HeadlessForm>
```

So here, as long as the custom component triggers a `focusout` event, validation would work ootb, no matter if there is no native form element, or what `name` it/they have or not have...

Btw, `triggerValidationFor` would _not_ call `e.stopPropagation()`, so the event would not stop to bubble at the `form.field` div level, if that was a concern. You could still attach event listeners to the `<form>` or even parent elements, and see the event...

NullVoxPopuli:
I, and our current users, also want direct access to the control elements 🙃

simonihmig:
> It's my understanding that you have no way to access the input component directly to add modifiers or attributes

No, that's not correct. Nothing in that regard would change with 3. here! See this example adding (imaginary) classes to _all_ elements:

```hbs
<HeadlessForm as |form|>
  <form.field @name="date" class="add-padding-to-wrapper" as |field|>  <-- this would now be a div that captures the event, and *knows* what @name it belongs to!
    <field.label class="align-right">Date:</field.label>
    <field.input class="border-gray"/>
    <field.errors class="text-red" />
  </form
</HeadlessForm>
```

In other words, there would still be no implicit element, that don't have any `...attributes` you can use to add whatever you want!

ember-headless-form/src/components/headless-form.ts
NullVoxPopuli: maybe this div is too much to assume. I imagine this may be something that a custom-control author would want to opt in to.

as in:
```hbs
<HeadlessForm as |form|>
  <form.field @name="date" as |field|> 
    <field.captureEvents @events={{"focusout"}}> or @events={{array "focusout" "change"}} ? idk
     ^-- this would now be a div that captures the event, and *knows* what @name it belongs to!
    <field.label>Date:</field.label>
      <CustomDatePickerComponentConsistingOfThreeSelectsWithUnknownNames
        @value={{field.value}}
        @onUpdate={{field.setValue}}
      />
    </field.captureEvents>
    <field.errors/>
  </form
</HeadlessForm>
```

But maybe that begins to defeat the purpose of the headless form?

>  would be curried into the component when yielding it as a contextual component, so that's a private implementation detail.

would there be a way to not add those event bindings if no validator is passed? (conditionally apply the on modifier?)

> Btw, triggerValidationFor would not call e.stopPropagation(), so the event would not stop to bubble at the form.field div level, if that was a concern

it was a concern! this is good news.

> You could still attach event listeners to the <form> or even parent elements, and see the event...

most excellent.

----------------------

I think I've now changed my opinion from being _all in_ on option 1, to _at least_ being a bit 50/50 on options 1 and 3 now.

I do have a concern though, so, what would happen if a field has multiple focusables? (this may sway me back to option 1? idk!)

```hbs
<HeadlessForm as |form|>
  <form.field @name="date" as |field|>
    <field.label>
       Date:
       <button {{on 'click' (fn this.helpAbout "date")>help</button>
        ^ - would we end up validating on events from here?
    </field.label>
    <CustomDatePickerComponentConsistingOfThreeSelectsWithUnknownNames
      @value={{field.value}}
      @onUpdate={{field.setValue}}
    />
    <field.errors/>
  </form
</HeadlessForm>
```

simonihmig:
Btw, I think in most cases you will want a wrapping div around the label and input anyway, like a flex container for positioning the children horizontally. So you could use that to do something like `<form.field class="flex" @name="foo">...`.

In our case, when we use `<Field>` (was that the name of the new reusable component? also didn't check the actual API, just a example here) as in 

```hbs
<HeadlessForm as |form|>
  <form.field @name="date" as |field|>  <-- this would now be a div that captures the event, and *knows* what @name it belongs to!
    <Field>
      <:label>
        <field.label class="align-right">Date:</field.label>
      </:label>
      <:input>    
        <field.input class="border-gray"/>
      </:input>
      <:hint>
        <field.errors class="text-red" />
      </:hint>
    </Field>
  </form.field>
</HeadlessForm>
```

we would probably end up with one additional wrapper div here. But that's also not the end of the world!? 🤷‍♂️

NullVoxPopuli:
having a built-in opportunity for controlling layout (form.field) is a good call out.

I think you sold me 🎉 

option 3 ✨

ynotdraw:
Thanks for the great write up! Yeah if that's not an issue at all then I think I'm good with option 3 as well! As you mention, we will want the label and input in a wrapping element anyway.

NullVoxPopuli:
If we all have Friday brain on Friday, we should all take the day off, right? 🙃

nicolechung:
Had to re-read all of this a few times, but option #3 sounds good to me now.

>Btw, I think in most cases you will want a wrapping div around the label and input anyway, like a flex container for positioning the children horizontally. So you could use that to do something like <form.field class="flex" @name="foo">....

This is good to know! Because sometimes we might want the label on the left or right (not just above)

> For 3., event.target.name would not matter, and everything would "just work"(TM)

I can't quite visualize how this would work until I see it in a demo or test / maybe I'm not following that whole thread here (quite likely, Friday brain for me too)

simonihmig:
> I can't quite visualize how this would work until I see it in a demo or test / maybe I'm not following that whole thread here (quite likely, Friday brain for me too)

I can certainly provide a working example, like I could amend this PR to switch to 3., shouldn't be much work. But the outline of the implementation is what I wrote in https://github.com/CrowdStrike/ember-headless-form/pull/24#discussion_r1089348895. See the first snippet there: when you have a div, the field can put on `{{on "focusout"}}` on it. Though it is not the form control that actually triggers the event, it will bubble up and the event listener attached to the div will see the event. And this event listener knows the name of the field, as `@name` is in scope for the `form.field` template.

> I imagine this may be something that a custom-control author would want to opt in to.

Oh, that's an interesting approach as well! 🤔  Option 4. I guess 😅

> would there be a way to not add those event bindings if no validator is passed? (conditionally apply the on modifier?)

Probably yes. Should be the case when you explicitly pass `@validate` functions. Not sure how exactly #10 will turn out, but maybe that will be difficult to support (conditional event handlers w/ native validation), as you apply things as plain HTML attributes like `required`, so we don't really know what is in use (unless we inspect DOM to change JS behaviour, which no, I don't think we want to do...)

> I do have a concern though, so, what would happen if a field has multiple focusables? 

Oh, another great point, I haven't thought about. So yes, without further measures, this would trigger validations as it is now. But there are ways to handle this. Off the top of my head:
* we could apply some heuristics on what elements we see as valid form controls. Like we would could check for `['button', 'a'].includes(event.target.tagName)` and ignore them if they match. However there might be a risk that some custom control is indeed a button (like a custom select), and focusing out is equivalent to focusing out of an input? 🤷‍♂️
* there is always the escape hatch that a custom control author could intercept the event and call `stopPropagation()`, if they think this shouldn't be handled by headless form. We could even provide that form the addon, like a `{{field.ignoreValidationEvents}}` modifier
* Saad mentioned this in a Zoom session recently: instead of showing validations on every focusout, we could only show them when the value has actually changed. I first thought of this as another mode for `@validateOn`, but it could also be the default behaviour for `@validateOn="blur"` maybe? Anyway, in this case just focussing out of the button in your example wouldn't do a thing.

Btw nitpick: we use `focusout` internally, but the API calls for `@validateOn="blur"`. Do want to keep this? The rationale would be that "blur" is more commonly used and understood, but I am not really sure that is the case!? 🤔

simonihmig:
I created a real PR (#25), but keeping this open here for further discussions.

I am leaning towards option 4 here, which is basically option 1 plus additional primitives for custom controls in case they are needed, like a yielded modifier (see above comment) as a variation of what @NullVoxPopuli [proposed with `field.captureEvents`](https://github.com/CrowdStrike/ember-headless-form/pull/24#discussion_r1089363380). WHat do you think?

PR #25, opened 2023-01-30
Support for `@validateOn`/`@revalidateOn` options

simonihmig:
As an iteration over #21, this adds the options to validate and revalidate (when initial validation had errors) not only on submit, but also on blur/focusout and change.

It supersedes #24, and still relies on option 1 discussed in that draft PR. In a future iteration it is expected that we had further primitives for custom controls, like "option 4" of yielding a modifier to capture focusout/change event, or yielding more validation state like `field.isValid`. But this can all be done on top of this PR.

### DIFF
--- a/ember-headless-form/src/components/-private/types.ts
+++ b/ember-headless-form/src/components/-private/types.ts
@@ -55,7 +55,7 @@ export type FieldValidateCallback<
  * Internal structure to track used fields
  * @private
  */
-export interface FieldData<
+export interface FieldRegistrationData<
   DATA extends FormData,
   KEY extends FormKey<DATA> = FormKey<DATA>
 > {
@@ -69,7 +69,7 @@ export interface FieldData<
 export type RegisterFieldCallback<
   DATA extends FormData,
   KEY extends FormKey<DATA> = FormKey<DATA>
-> = (name: KEY, field: FieldData<DATA, KEY>) => void;
+> = (name: KEY, field: FieldRegistrationData<DATA, KEY>) => void;
 
 export type UnregisterFieldCallback<
   DATA extends FormData,
@@ -79,6 +79,4 @@ export type UnregisterFieldCallback<
 /**
  * Mapper type to construct subset of objects, whose keys are only strings (and not number or symbol)
  */
-export type OnlyStringKeys<T extends object> = {
-  [P in Extract<keyof T, string>]: T[P];
-};
+export type OnlyStringKeys<T extends object> = Pick<T, keyof T & string>;
--- a/ember-headless-form/src/components/headless-form.ts
+++ b/ember-headless-form/src/components/headless-form.ts
@@ -89,18 +89,26 @@ export default class HeadlessFormComponent<
     const { validateOn } = this;
 
     return validateOn === 'submit'
-      ? undefined
-      : validateOn === 'blur'
+      ? // no need for dynamic validation, as validation always happens on submit
+        undefined
+      : // our component API expects "blur", but the actual blur event does not bubble up, so we use focusout internally instead
+      validateOn === 'blur'
       ? 'focusout'
       : validateOn;
   }
 
   get fieldRevalidationEvent(): 'focusout' | 'change' | undefined {
-    const { revalidateOn } = this;
+    const { validateOn, revalidateOn } = this;
 
     return revalidateOn === 'submit'
+      ? // no need for dynamic validation, as validation always happens on submit
+        undefined
+      : // when validation happens more frequently than revalidation, then we can ignore revalidation, because the validation handler will already cover us
+      validateOn === 'change' ||
+        (validateOn === 'blur' && revalidateOn === 'blur')
       ? undefined
-      : revalidateOn === 'blur'
+      : // our component API expects "blur", but the actual blur event does not bubble up, so we use focusout internally instead
+      revalidateOn === 'blur'
       ? 'focusout'
       : revalidateOn;
   }
--- a/ember-headless-form/src/components/headless-form.ts
+++ b/ember-headless-form/src/components/headless-form.ts
@@ -48,6 +48,9 @@ export interface HeadlessFormComponentSignature<DATA extends UserData> {
   };
 }
 
+/**
+ * This internal data structure maintains information about each field that is registered to the form by `registerField`.
+ */
 class FieldData<
   DATA extends FormData,
   KEY extends FormKey<DATA> = FormKey<DATA>
@@ -56,8 +59,14 @@ class FieldData<
     this.validate = fieldRegistration.validate;
   }
 
+  /**
+   * tracked state that enabled a dynamic validation of a field *before* the whole form is submitted, e.g. by `@validateOn="blur" and the blur event being triggered for that particular field.
+   */
   @tracked validationEnabled = false;
 
+  /**
+   * The *field* level validation callback passed to the field as in `<form.field @name="foo" @validate={{this.validateCallback}}>`
+   */
   validate?: FieldValidateCallback<DATA, KEY>;
 }
 
@@ -70,11 +79,21 @@ export default class HeadlessFormComponent<
   // we cannot use (modifier "on") directly in the template due to https://github.com/emberjs/ember.js/issues/19869
   on = on;
 
+  /**
+   * A copy of the passed `@data` stored internally, which is only passed back to the component consumer after a (successful) form submission.
+   */
   internalData: DATA = new TrackedObject(this.args.data ?? {}) as DATA;
 
   fields = new Map<FormKey<FormData<DATA>>, FieldData<FormData<DATA>>>();
 
+  /**
+   * The last result of calling `this.validate()`.
+   */
   @tracked lastValidationResult?: ErrorRecord<FormData<DATA>>;
+
+  /**
+   * When this is set to true by submitting the form, eventual validation errors are show for *all* field, regardless of their individual dynamic validation status in `FieldData#validationEnabled`
+   */
   @tracked showAllValidations = false;
 
   get validateOn(): ValidateOn {
@@ -85,6 +104,9 @@ export default class HeadlessFormComponent<
     return this.args.revalidateOn ?? 'change';
   }
 
+  /**
+   * Return the event type that will be listened on for dynamic validation (i.e. *before* submitting)
+   */
   get fieldValidationEvent(): 'focusout' | 'change' | undefined {
     const { validateOn } = this;
 
@@ -97,6 +119,9 @@ export default class HeadlessFormComponent<
       : validateOn;
   }
 
+  /**
+   * Return the event type that will be listened on for dynamic *re*validation, i.e. updating the validation status of a field that has been previously marked as invalid
+   */
   get fieldRevalidationEvent(): 'focusout' | 'change' | undefined {
     const { validateOn, revalidateOn } = this;
 
@@ -113,6 +138,9 @@ export default class HeadlessFormComponent<
       : revalidateOn;
   }
 
+  /**
+   * Return true if validation has happened (by submitting or by an `@validateOn` event being triggered) and at least one field is invalid
+   */
   get hasValidationErrors(): boolean {
     // Only consider validation errors for which we actually have a field rendered
     return this.lastValidationResult
@@ -122,6 +150,9 @@ export default class HeadlessFormComponent<
       : false;
   }
 
+  /**
+   * Call the passed validation callbacks, defined both on the whole form as well as on field level, and return the merged result for all fields.
+   */
   async validate(): Promise<ErrorRecord<FormData<DATA>> | undefined> {
     let errors: ErrorRecord<FormData<DATA>> | undefined = undefined;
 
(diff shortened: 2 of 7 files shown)


ember-headless-form/src/components/headless-form.ts
```diff
@@ -146,4 +219,36 @@ export default class HeadlessFormComponent<
...
+
+    if (name) {
+      const field = this.fields.get(name as FormKey<FormData<DATA>>);
+
+      if (field) {
+        this.lastValidationResult = await this.validate();
+        field.validationEnabled = true;
+      }
+    } else {
+      // @todo how to handle custom controls that don't emit focusout/change events from native form controls?
```
simonihmig, line 236: The DX should be improved here by emitting a warning, by a future PR (see PR description) that provides additional primitives for better custom controls support, and mentions these primitives in the warning. See keeping this as-is for now...

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+        );
+      });
+
+      test('validation errors are exposed as field.errors on blur', async function (assert) {
+        const data: TestFormData = { firstName: 'Tony', lastName: 'Foo' };
+
+        await render(<template>
+          <HeadlessForm
+            @data={{data}}
+            @validateOn="blur"
```
nicolechung, line 949: Question: in glint does this auto hint `blur` or `change`?

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+            </form.field>
+            <button type="submit" data-test-submit>Submit</button>
+          </HeadlessForm>
+        </template>);
+
+        await fillIn('[data-test-first-name]', 'Foo');
+
+        assert.false(
+          validateCallback.called,
+          '@validate is not called while typing'
```
nicolechung, line 1167: ✅  nice!

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+        assert.true(
+          validateCallback.getCall(1).calledWith('Tony', 'firstName', {
+            ...data,
+            firstName: 'Tony',
+          }),
+          '@validate is called with form data'
+        );
+      });
+
+      test('validation errors are exposed as field.errors on blur', async function (assert) {
```
nicolechung, line 1281: Suggestion: `'validation errors are exposed as field.errors on blur'` -> `'validation errors are exposed as field.errors on @revalidateOn="blur'`

nicolechung (commented):
So far looking GREAT 💯 ...I have a bunch of questions / suggestions...but very excited to see how easy this is to use

ember-headless-form/src/components/headless-form.ts
```diff
@@ -146,4 +219,36 @@ export default class HeadlessFormComponent<
...
+  }
+
+  @action
+  async handleFieldRevalidation(e: Event): Promise<void> {
+    const { target } = e;
+    const { name } = target as HTMLInputElement;
+
+    if (name) {
+      if (this.showErrorsFor(name as FormKey<FormData<DATA>>)) {
+        this.lastValidationResult = await this.validate();
```
NullVoxPopuli, line 247: interacting with `this` after an await isn't safe 🙈 

I keep meaning to work on [this lint](https://github.com/ember-cli/eslint-plugin-ember/pull/1421) but... priorities 🙃 

be sure to `if (isDestroyed(this) || isDestroying(this)) return` after an await

ember-headless-form/src/components/headless-form.ts
```diff
@@ -67,6 +85,34 @@ export default class HeadlessFormComponent<
...
+    return validateOn === 'submit'
+      ? // no need for dynamic validation, as validation always happens on submit
+        undefined
+      : // our component API expects "blur", but the actual blur event does not bubble up, so we use focusout internally instead
+      validateOn === 'blur'
+      ? 'focusout'
+      : validateOn;
+  }
+
+  get fieldRevalidationEvent(): 'focusout' | 'change' | undefined {
```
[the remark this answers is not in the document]
simonihmig: Added, and a bunch of more JSDoc annotations along that!

This was good to call out, thank you! Indeed there are different things playing together here, which were clear to me as I was working on the things directly for many hours, but which might not be obvious on a review or when working in the codebase and not being super familiar with it. So, yeah, I must remind myself to explain things better here!

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+        assert.false(
+          validateCallback.called,
+          '@validate is not called while typing'
+        );
+
+        await blur('[data-test-first-name]');
+
+        assert.true(
+          validateCallback.calledWith({ ...data, firstName: 'Foo' }),
+          '@validate is called with form data'
```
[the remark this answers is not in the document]
simonihmig: Applied most of your assertion suggestions here, and at all the other places were similar things were tested! Thanks!

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: Yeah, so I just tested this, and Glint calls out an error if you pass anything but the expected values, so here `blur`, `change` or `submit`. But it actually did not provide something like autocomplete (like typing `b` and suggesting `blur`), which I was expecting. I am not sure why that is, or if I am expecting too much 🤔 

But at least the types seem correct from our end seem correct...

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+
+        assert.true(
+          validateCallback.calledTwice,
+          '@validate is called for revalidation on blur'
+        );
+        assert.true(
+          validateCallback
+            .getCall(1)
+            .calledWith({ ...data, firstName: 'Tony' }),
+          '@validate is called with form data'
```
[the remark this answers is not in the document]
simonihmig: Yeah, that is what is entered basically, which updates on every change. 

And no, it is _not_ the native [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) API, just a simple POJO. But that's deliberately the case, to support custom control use cases that we touched already, like the 3-select-datepicker, which should map to a single `date` field instead of 3 individual ones, which is what a native `new FormData(formElement)` would give you.

If there comes up a need to actually provide a `FormData` instance instead of that POJO, we can do that easily by creating it from that POJO. But I think we should keep the functionality to have form data that does not map 1:1 to native form controls.

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: I would disagree here, because
* we have that wrapped in a `module()` that already makes clear that we are testing the `@revalidateOn="blur'` scenario
* the `on blur` refers to what the user is doing here, not what the dev is passing to the component

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+        assert
+          .dom('[data-test-last-name-errors]')
+          .doesNotExist(
+            'validation errors are not rendered for untouched fields'
+          );
+      });
+    });
+  });
+
+  module(`@revalidateOn`, function () {
```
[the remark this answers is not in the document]
simonihmig: I have mixed feelings about this, because it is hard to draw the boundaries here for what should go into what testing file, as all these things belong together and can be combined. For example I don't have test coverage here for all the combinations, like `@validateOn="blur" @revalidateOn="change"`. But this would be a good and valid test scenario, and into which test file should be put it then?

Also we will probably add additional validation-related features, like this `captureEvents` modifier (aka "option 4"), which we might want to test with a particular `@revalidateOn` config. So should this then be in the `validation` file, the `revalidate` file or its own? 

Idk, like I am not opposed to this, as I said I am just unsure how to draw the boundaries and not cause more confusion than we wanted to prevent by splitting this up...

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+        );
+        assert.true(
+          validateCallback
+            .getCall(1)
+            .calledWith({ ...data, firstName: 'Tony' }),
+          '@validate is called with form data'
+        );
+      });
+
+      test('field validation callback is called on blur', async function (assert) {
```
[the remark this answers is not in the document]
simonihmig: see above

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -830,4 +845,778 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+
+        await render(<template>
+          <HeadlessForm
+            @data={{data}}
+            @revalidateOn="change"
+            @validate={{validateFormCallbackSync}}
+            as |form|
+          >
+            <form.field @name="firstName" as |field|>
+              <field.label>First Name</field.label>
```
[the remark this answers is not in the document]
simonihmig: Great suggestion! Yes, we should do that, I'll keep that in my mind (hopefully, scold me when I forget :wink:). But I would add that in a future PR that also adds the primitive to support that use case, which we currently don't have yet.

ember-headless-form/src/components/-private/types.ts
```diff
@@ -55,7 +55,7 @@ export type FieldValidateCallback<
  * Internal structure to track used fields
  * @private
  */
-export interface FieldData<
+export interface FieldRegistrationData<
```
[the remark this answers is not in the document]
simonihmig: I was splitting this up into two concerns, that's why I renamed this.

So there is information passed form the field component to the parent form component via the `registerField` action. Basically the field register itself with its parent form, so the form has data available that the dev passed to the field. Currently this is just the _field_ level validation callback, which the form needs to have available. This is what the `FieldRegistrationData` contains.

Then in this PR there is a new [`FieldData` class being added](https://github.com/CrowdStrike/ember-headless-form/pull/25/files#diff-71dca24e838c62279dc22478137e692716eddbccaf782f3901bfda818c5e67daR54-R71), which gets created when registering the field, so it uses `FieldRegistrationData`, but it also contains more than that (the `validationEnabled` tracked property), which the field does not need to care about.

So that's why we have these two different but related concepts now.

As mentioned earlier, I added a bunch of JSDoc comments, also here, so hopefully that becomes more clear now!

ember-headless-form/src/components/headless-form.ts
simonihmig: Good catch! So I can certainly add `isDestroy{ing,ed}` guards here, even if it's just for code style or best practices.

But I'd first like to understand why this could be a problem here, i.e. what is unsafe about it? 

So when the async action still has a pointer to `this` while the component that `this` refers to is already "destroyed" (unmounted), then `this` will still exist. There is no way it could be "gone", because that's how Garbage Collection works: here our async action is basically the retainer for that object preventing GC.

So it exists, but it could be in a "destroyed" state. In classic Ember, a call to `set(this,...)` or `this.set()` would fail, so yes, that would be bad. But in Octane, would mutating a tracked property be so bad? I don't think anything will throw here, right? And when our form is already "unmounted" (in React's terminology, do we have a better term, unrendered?), this tracked property will just not be consumed anymore, so also not cause rerenders, but that should be all, no further harm caused, or am I missing something?

Compare that to [this comment](https://github.com/ember-cli/eslint-plugin-ember/issues/418#issuecomment-1009181665) here on the linting issue...

Loosely related: should we wrap the async methods (that call potentially async validators) with `waitFor()` from `@ember/test-waiters`, so there are able to cover async things happening that Ember does not auto-detect as async? I would think so, yes?

ynotdraw (approved):
Code looks great to me! I'm approving, but also curious about the `isDestroy{ing,ed}` comments above so going to track those.

ember-headless-form/src/components/-private/types.ts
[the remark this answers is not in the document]
nicolechung: Thanks for this!

test-app/tests/integration/components/headless-form-validation-test.gts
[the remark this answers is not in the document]
nicolechung: Sounds good to me, let's leave it as is

ember-headless-form/src/components/headless-form.ts
NullVoxPopuli: >  I would think so, yes?

yeah, that's a good way to solve this whole class of problem.

> what is unsafe about it?

you raise some good points, and I'm going to have to do some concrete research with shareable examples before continuing on with the lint 😅

simonihmig: > accessing services that is the problem in this situation, since service instantiation is lazy

Oh right, that can be problematic indeed!

simonihmig: So I added `@waitFor` to cover async validations, and made the test cover async validations properly (i.e. they were indeed failing without that `@waitFor`), see [this commit](https://github.com/CrowdStrike/ember-headless-form/pull/25/commits/0fb2cec919c1b5a85124c6bb01e2e9e3d89dae5e)

Regarding the `isDestroy{ing,ed}` topic, should we address this here and now, or eventually iterate on it when it's more clear how to address this?

In other words: is this good to merge? @NullVoxPopuli @nicolechung

NullVoxPopuli: we're good to merge. 

isDetroy{ing,ed}, I think I want to iterate on the lint, so we can have a definite error case to match on, and then provide an autofix as well

NullVoxPopuli (approved):

nicolechung (approved):

PR #29, opened 2023-02-06
Support native validation

### LINKED ISSUE #10: Support for native HTML5 validation API

simonihmig:
Picks up any validation errors from the native [HTML5 constraint API](https://developer.mozilla.org/en-US/docs/Web/HTML/Constraint_validation), and makes them available for rendering the same way as existing custom validations. That way at least simple validations (like `required`) can be implemented without any JavaScript-based validations, or can be _combined_ with them. 

Closes #10

### DIFF
--- a/ember-headless-form/src/components/headless-form.hbs
+++ b/ember-headless-form/src/components/headless-form.hbs
@@ -1,5 +1,7 @@
 <form
+  novalidate
   ...attributes
+  {{this.registerForm}}
   {{on 'submit' this.onSubmit}}
   {{(if
     this.fieldValidationEvent
--- a/ember-headless-form/src/components/headless-form.ts
+++ b/ember-headless-form/src/components/headless-form.ts
@@ -5,9 +5,11 @@ import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 import { waitFor } from '@ember/test-waiters';
 
+import { modifier } from 'ember-modifier';
 import { TrackedObject } from 'tracked-built-ins';
 
 import FieldComponent from './-private/field';
+import { mergeErrorRecord } from './-private/utils';
 
 import type { HeadlessFormFieldComponentSignature } from './-private/field';
 import type {
@@ -20,7 +22,11 @@ import type {
   UserData,
   ValidationError,
 } from './-private/types';
-import type { ComponentLike, WithBoundArgs } from '@glint/template';
+import type {
+  ComponentLike,
+  ModifierLike,
+  WithBoundArgs,
+} from '@glint/template';
 
 type ValidateOn = 'change' | 'blur' | 'submit';
 
@@ -86,6 +92,12 @@ export default class HeadlessFormComponent<
 
   // we cannot use (modifier "on") directly in the template due to https://github.com/emberjs/ember.js/issues/19869
   on = on;
+
+  formElement?: HTMLFormElement;
+
+  registerForm = modifier((el: HTMLFormElement, _p: []) => {
+    this.formElement = el;
+  }) as unknown as ModifierLike<unknown>; // @todo getting Glint errors without this. Try again with Glint 1.0 (beta)!
 
   /**
    * A copy of the passed `@data` stored internally, which is only passed back to the component consumer after a (successful) form submission.
@@ -162,34 +174,82 @@ export default class HeadlessFormComponent<
    * Call the passed validation callbacks, defined both on the whole form as well as on field level, and return the merged result for all fields.
    */
   @waitFor
-  async validate(): Promise<ErrorRecord<FormData<DATA>> | undefined> {
-    let errors: ErrorRecord<FormData<DATA>> | undefined = undefined;
-
-    if (this.args.validate) {
-      errors = await this.args.validate(this.internalData);
-    }
-
-    if (!errors) {
-      errors = {};
-    }
+  async validate(): Promise<ErrorRecord<FormData<DATA>>> {
+    const nativeValidation = this.validateNative();
+    const customFormValidation = await this.args.validate?.(this.internalData);
+    const customFieldValidations: ErrorRecord<FormData<DATA>>[] = [];
 
     for (const [name, field] of this.fields) {
-      const fieldValidation = await field.validate?.(
+      const fieldValidationResult = await field.validate?.(
         this.internalData[name],
         name,
         this.internalData
       );
 
-      if (fieldValidation) {
-        const existingFieldErrors = errors[name];
-
-        errors[name] = existingFieldErrors
-          ? [...existingFieldErrors, ...fieldValidation]
-          : fieldValidation;
+      if (fieldValidationResult) {
+        customFieldValidations.push({
+          [name]: fieldValidationResult,
+        } as ErrorRecord<FormData<DATA>>);
       }
     }
 
-    return Object.keys(errors).length > 0 ? errors : undefined;
+    return mergeErrorRecord(
+      nativeValidation,
+      customFormValidation,
+      ...customFieldValidations
+    );
+  }
+
+  validateNative(): ErrorRecord<FormData<DATA>> | undefined {
+    const form = this.formElement;
+
+    assert(
+      'Form element expected to be present. If you see this, please report it as a bug to ember-headless-form!',
+      form
+    );
+
+    if (form.checkValidity()) {
+      return;
+    }
+
+    const errors: ErrorRecord<FormData<DATA>> = {};
+
+    for (const el of form.elements) {
+      // This is just to make TS happy, as we need to access properties on el that only form elements have, but elements in `form.elements` are just typed as plain `Element`. Should never occur in reality.
+      assert(
+        'Unexpected form element. If you see this, please report it as a bug to ember-headless-form!',
+        el instanceof HTMLInputElement ||
+          el instanceof HTMLTextAreaElement ||
+          el instanceof HTMLSelectElement ||
+          el instanceof HTMLButtonElement ||
+          el instanceof HTMLFieldSetElement ||
+          el instanceof HTMLObjectElement ||
+          el instanceof HTMLOutputElement
+      );
+
+      if (el.validity.valid) {
+        continue;
+      }
+
+      const name = el.name as FormKey<FormData<DATA>>;
+
+      if (this.fields.has(name)) {
+        errors[name] = [
+          {
+            type: 'native',
+            value: this.internalData[name],
+            message: el.validationMessage,
+          },
+        ];
+      } else {
+        warn(
+          `An invalid form element with name "${name}" was detected, but this name is not used as a form field. It will be ignored for validation. Make sure to apply the correct name to custom form elements that participate in form validation!`,
+          { id: 'headless-form.invalid-control-for-unknown-field' }
+        );
+      }
+    }
+
+    return errors;
   }
 
   /**
--- a/test-app/app/templates/index.hbs
+++ b/test-app/app/templates/index.hbs
@@ -6,7 +6,8 @@
 >
   <form.field @name='name' as |field|>
     <field.label>Name</field.label>
-    <field.input />
+    <field.input required />
+    <field.errors />
   </form.field>
 
   <button type='submit'>Submit</button>
--- a/test-app/tests/integration/components/headless-form-basic-test.gts
+++ b/test-app/tests/integration/components/headless-form-basic-test.gts
@@ -25,13 +25,19 @@ module('Integration Component HeadlessForm > Basics', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it renders form markup', async function (assert) {
-    await render(<template><HeadlessForm class="foo" novalidate /></template>);
+    await render(<template>
+      <HeadlessForm class="foo" autocomplete="off" />
+    </template>);
 
     assert.dom('form').exists('it renders as <form>');
     assert
       .dom('form')
       .hasClass('foo', 'it accepts custom HTML classes')
-      .hasAttribute('novalidate', '', 'it accepts arbitrary HTML attributes');
+      .hasAttribute(
+        'autocomplete',
+        'off',
+        'it accepts arbitrary HTML attributes'
+      );
   });
 
   module('form.field', function () {
--- a/test-app/tests/integration/components/headless-form-native-validation-test.gts
+++ b/test-app/tests/integration/components/headless-form-native-validation-test.gts
@@ -1,0 +1,754 @@
+/* eslint-disable no-undef -- Until https://github.com/ember-cli/eslint-plugin-ember/issues/1747 is resolved... */
+/* eslint-disable simple-import-sort/imports,padding-line-between-statements,decorator-position/decorator-position -- Can't fix these manually, without --fix working in .gts */
+
+import { blur, click, fillIn, render } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+
+import HeadlessForm from 'ember-headless-form/components/headless-form';
+import sinon from 'sinon';
+import { setupRenderingTest } from 'test-app/tests/helpers';
+
+import type { RenderingTestContext } from '@ember/test-helpers';
+
+module(
+  'Integration Component HeadlessForm > Native validation',
+  function (hooks) {
+    setupRenderingTest(hooks);
+
+    interface TestFormData {
+      firstName?: string;
+      lastName?: string;
+    }
+
+    test('form has novalidate to render custom validation errors', async function (assert) {
+      const data: TestFormData = {};
+      const submitHandler = sinon.spy();
+
+      await render(<template>
+        <HeadlessForm @data={{data}} @onSubmit={{submitHandler}} as |form|>
+          <form.field @name="firstName" as |field|>
+            <field.label>First Name</field.label>
+            <field.input required data-test-first-name />
+          </form.field>
+          <button type="submit" data-test-submit>Submit</button>
+        </HeadlessForm>
+      </template>);
+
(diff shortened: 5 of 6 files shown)


test-app/tests/integration/components/headless-form-basic-test.gts
```diff
@@ -25,13 +25,19 @@ module('Integration Component HeadlessForm > Basics', function (hooks) {
...
-    await render(<template><HeadlessForm class="foo" novalidate /></template>);
+    await render(<template>
+      <HeadlessForm class="foo" autocomplete="off" />
+    </template>);
 
     assert.dom('form').exists('it renders as <form>');
     assert
       .dom('form')
       .hasClass('foo', 'it accepts custom HTML classes')
-      .hasAttribute('novalidate', '', 'it accepts arbitrary HTML attributes');
```
simonihmig, line 34: Removed the example using `novalidate` here, because the form now has this _by default_. So testing for any other HTML attribute instead, to not have a possibly false positive here.

ember-headless-form/src/components/headless-form.hbs
```diff
@@ -1,5 +1,7 @@
 <form
+  novalidate
```
simonihmig, line 2: This now always suppresses the native way of rendering validation errors, as in:

image

Instead errors are rendered via the existing yielded `field.errors` using any possible custom markup.

Eventually in a future iteration, we might also allow opting into a most simple native-only mode?

NullVoxPopuli: >  into a most simple native-only mode?

yeah, I think this is pretty important to support. We want to use _T H E  P L A T F O R M_, as much as possible, and while native validation isn't great (dare I said bad? 🙈 ), we're somewhat "breaking the web" by forcing an opt-out.

_instead_, can we only add `novalidate` conditionally? like, is there some way to know that we want to use custom errors?.
Is this worth the trade-offs? maybe an flag to `@nativeErrors` or something? idk

test-app/app/templates/index.hbs
```diff
@@ -6,7 +6,8 @@
 >
   <form.field @name='name' as |field|>
```
NullVoxPopuli, line 7: I love this demo <3

ember-headless-form/src/components/headless-form.hbs
NullVoxPopuli: (i'm not 100% on my own opinion here, as I think what you have implemented is _the happy path_)

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+          </form.field>
+          <button type="submit" data-test-submit>Submit</button>
+        </HeadlessForm>
+      </template>);
+
+      await click('[data-test-submit]');
+
+      assert.true(
+        invalidHandler.calledOnce,
+        '@onInvalid was called when required field in empty'
```
NullVoxPopuli, line 80: what arguments does it receive? should those be asserted here as well?

test-app/app/templates/index.hbs
nicolechung: Not really related, but where does this template get rendered? Because in integration tests, they all seem to render their own templates?

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+          <button type="submit" data-test-submit>Submit</button>
+        </HeadlessForm>
+      </template>);
+
+      await click('[data-test-submit]');
+
+      assert
+        .dom('[data-test-first-name-errors]')
+        .exists({ count: 1 })
+        .hasAnyText(); // validation error message is browser and locale dependant, so testing against actual message would be very brittle.
```
NullVoxPopuli, line 170: I appreciate the comment here!  <3

ember-headless-form/src/components/headless-form.hbs
nicolechung: Can it be overridden with the `...attributes`? Like if someone adds `novalidate=false`?

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+        .dom('[data-test-last-name-errors]')
+        .doesNotExist(
+          'validation errors are not rendered before validation happens'
+        );
+
+      await click('[data-test-submit]');
+
+      assert
+        .dom('[data-test-first-name-errors]')
+        .exists({ count: 1 }, 'validation errors appear when validation fails');
```
nicolechung, line 143: I like that this test is in here [based on reading this comment](https://github.com/CrowdStrike/ember-headless-form/pull/29/files/a253d4230b82f6097c10e4299d7013e42351a3fe#r1097548956)

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+            .dom('[data-test-last-name-errors]')
+            .doesNotExist(
+              'validation errors are not rendered for untouched fields'
+            );
+        });
+      });
+    });
+
+    module(`@revalidateOn`, function () {
+      module('@revalidateOn=blur', function () {
```
NullVoxPopuli, line 546: this arg is super cool, and I'm really excited that it exists 🎉

NullVoxPopuli (approved):
lookin good! thanks for the thorough tests!  (and having them in a separate file!)

ember-headless-form/src/components/headless-form.ts
```diff
@@ -162,34 +174,82 @@ export default class HeadlessFormComponent<
...
+        el instanceof HTMLInputElement ||
+          el instanceof HTMLTextAreaElement ||
+          el instanceof HTMLSelectElement ||
+          el instanceof HTMLButtonElement ||
+          el instanceof HTMLFieldSetElement ||
+          el instanceof HTMLObjectElement ||
+          el instanceof HTMLOutputElement
+      );
+
+      if (el.validity.valid) {
```
ynotdraw, line 230: I actually quite like the native validation API!

ynotdraw (approved):
So awesome to see this already! 🎉 Excellent work

ember-headless-form/src/components/headless-form.hbs
sukima: With [ember-validity-modifier](https://github.com/sukima/ember-validity-modifier#example-validation-on-form-submit) I provide the user with a `verify-form-validity` modifier which will handle both the `novalidate` and `checkValidity()` versus `reportValidity()` checks.

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+          assert
+            .dom('[data-test-first-name-errors]')
+            .exists(
+              { count: 1 },
+              'validation errors appear on blur when validation fails'
+            );
+          assert
+            .dom('[data-test-last-name-errors]')
+            .doesNotExist(
+              'validation errors are not rendered for untouched fields'
```
nicolechung, line 539: Thank you for testing for this!

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+          '[data-test-first-name-errors] [data-test-error="2"] [data-test-error-value]'
+        )
+        .hasText('foo123');
+      assert
+        .dom(
+          '[data-test-first-name-errors] [data-test-error="2"] [data-test-error-message]'
+        )
+        .hasText('Foo is an invalid first name!');
+
+      assert.dom('[data-test-last-name-errors]').doesNotExist();
```
nicolechung, line 417: Suggestion, nit: this test only seems to test for `errors` and not the success state?

nicolechung (approved):
Thank you for the very descriptive and detailed tests...really helps me understand how this library works.

ember-headless-form/src/components/headless-form.hbs
nicolechung: Oh interesting! So adding `{{verify-form-validity}}` automatically adds `novalidate`? https://github.com/sukima/ember-validity-modifier/blob/52471db575ec15b852286e3572a94278e8cfe5c4/tests/integration/modifiers/verify-form-validity-test.js#L39

simonihmig:
Thanks for reviewing so quickly, will go through all of your comments tomorrow morning!

ember-headless-form/src/components/headless-form.hbs
simonihmig: > instead, can we only add novalidate conditionally? 

I think we need to add `novalidate` _always_, as our JS backed form needs to be in control when to render validations.  It seems `{{verify-form-validity}}` is also [doing that unconditionally](https://github.com/sukima/ember-validity-modifier/blob/main/addon/-private/validity.js#L163). But that does not mean we cannot do native validation renderings, see below...

> like, is there some way to know that we want to use custom errors?.

Maybe, we could establish some additional "registration protocol" between the field and its yielded `field.errors`, so when `field.errors` is used, that's a signal that users want custom error rendering. However `field.errors` is per field, and the validation mode is for the whole form. Also we might yield also the raw errors array also in addition to the `field.errors` _component_, in which case that's impossible to see if that is used. At the end I would think more of an additional mode you opt-in, I like the `@reportValidity={{true}}` approach that `{{verify-form-validity}}` offers in a similar way.

This matches what we have to do technically as @sukima pointed out: instead of `checkValidity()` (which we currently use) we would have to call `reportValidity()` to trigger the browser to actually show the errors in its native way. In addition to that though, we also have to account for 👇

> Can it be overridden with the ...attributes? Like if someone adds novalidate=false?

Not really, because that would trigger _both_ native rendering as well as our custom rendering (tested it yesterday before our discussion here):

image

See the same (German, sorry) validation rendered both in native and custom way

So as said above, we need to call `reportValidity()` and suppress yielding our custom errors, which again calls for a dedicated mode like `@reportValidity={{true}}`.

Things start to get more complicated though in that mode, as the question comes up what to do with the JS-based validation callbacks that you can _also_ use to validate fields (`@validate={{validateFn}}` implemented in #21), show they also render in the native way? Which would mean we would have to "register" those errors with the native constraint API using `setCustomValidity()`, which is kinda the opposite way of what we do here (setting native errors from JS, instead of querying native errors from JS)

Given all of that...

> I have yet to see a situation where a designer or a product owner WANTS that native error style

... and that I agree with that, and the need to get something going that at least serves _our current needs_, I was hoping we don't have to bother with this, or at least not _now_. If the need arises, I think we should be able to provide a non-breaking opt-in mode for that.

simonihmig: Btw, thanks @sukima for chiming in here! I actually wanted to ping you anyway, given that I was well aware of your work on `ember-validity-modifier`, even before joining CS and getting assigned to this forms project. 

It actually inspired this work, not the actual APIs so much, but more the concept of not doing things very specific to Ember, but leveraging _THE PLATFORM_ (as Preston calls it 🙂). Speaking of the API, this is doing a more component-focused approach vs. the modifier approach you took. The reason for this is that this addon addresses basically two major concerns: validations and accessible markup by default. So it is not truly headless (no idea what the actual definition for this is) in that it does not render _anything_, but it does render things that it has strong opinions on, like again accessible markup (inputs connected to labels, aria attributes etc.), but it makes no assumptions on top of that (no styling, no classes etc.).

By using the Platform, hopefully it is able to compose well with other solutions. E.g. it would nice to see if we could also incorporate custom validation rules just by applying your `{{validity}}` modifier on a headless-form input. There is [this test](https://github.com/CrowdStrike/ember-headless-form/pull/29/files#diff-e54bb85c25006594b458dfc882cd3ab24ddaae50f33ab73705b8200ee394709dR225-R298) with this scenario in mind, so it could already work (though I am not entirely sure how [this](https://github.com/sukima/ember-validity-modifier/blob/main/addon/-private/validity.js#L130-L131) would play out)

So long story short, please share your thoughts here as much as possible and as your time permits. Happy to also have a more high-bandwidth conversation (Zoom) if you want!

test-app/app/templates/index.hbs
simonihmig: Yeah, it was just for seeing something on `htttp://localhost:4200/`, not used in tests at all. With #4 we will probably kill this, and add more feature-complete kitchen-sink like demos there!

test-app/tests/integration/components/headless-form-native-validation-test.gts
simonihmig: Good catch, yeah, I was a bit sloppy here 😅

This receives the same arguments as for custom validations like in [this test](https://github.com/CrowdStrike/ember-headless-form/blob/main/test-app/tests/integration/components/headless-form-validation-test.gts#L196-L207), but the thing is as mentioned elsewhere that the actual validation message is brittle, so I'd test for the _existence_ of it, but not the actual value. But I was not sure how to do this easily with sinon. Any suggestion?

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -0,0 +1,754 @@
...
+      assert
+        .dom(
+          '[data-test-first-name-errors] [data-test-error="0"] [data-test-error-value]'
+        )
+        .hasText('foo123');
+      assert
+        .dom(
+          '[data-test-first-name-errors] [data-test-error="0"] [data-test-error-message]'
+        )
+        .hasAnyText(); // validation error message is browser and locale dependant, so testing against actual message would be very brittle.
```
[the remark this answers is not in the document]
simonihmig: Yeah, there is also btw some kind of DX/UX tradeoff here that we (and UX designers/writers) need to be aware of: while just adding e.g. `required` somewhere is super easy instead of having to pass custom JS validation logic (be they custom functions, or ember-changeset, or yup or whatnot in the future), we do loose the capability to influence the actual wording here!

test-app/tests/integration/components/headless-form-native-validation-test.gts
simonihmig: Good point, I'll add another one for this!

test-app/app/templates/index.hbs
nicolechung: Thanks for the explanation!

test-app/tests/integration/components/headless-form-native-validation-test.gts
NullVoxPopuli: if it's tested elsewhere, that should be fine 💯

ember-headless-form/src/components/headless-form.hbs
sukima: I had to do a lot more logic workarounds to manage my validity modifier because the components I was using them on lacked `...attributes` which meant the modifier was no longer attached to a specific form field but to a wrapping div and used event bubbling to capture the input/blur/etc. events. Since the _platform_ is taking on the responsibility to manage the `<form>` itself and the form `<input>` fields within providing a modifier for validations would simply be syntactical sugar IMHO.

On to native validity. It isn't reactive in the same way Ember tracked properties are thus you would need an action to trap the causes for the validity to change. Luckily in order to support custom validity you have to do that anyways. My addon did this by trapping the input/change/blur events. But because of bubbling you can do this do this in a wrapping div/component.

To manage the checks and reporting the `<form>` itself keeps a dynamic tally of field elements and can easily provide an iterator for them.

I'm brainstorming here but this is how I might design a system for this. Internally the form component exposes an internal class that tracks validation messages. Attached to the `<form>` by means of an `{{on "validate" …}}` modifier. This form component also traps the `submit` event and based on a flag with either `checkValidity()` or `reportValidity()` before calling the `@onSubmit` action.

Then each field component takes a `@validity` argument which is an array (or singular) of validation functions to call on the inputs’ input/change/blur events. You could leverage a modifier in this case allowing more flexibility in the positional/named params if desired.

In the end ember-validity-modifier can be used _internally_ for the plumbing or you can implement on its own. Either way I think the original event based management is still worth leveraging. Specifically this communication flow:

diagram of form validation communication flow

In the above figure, the Vanilla™ validity and the Ember™ validity are tracked separately but linked by the side effect of the `validated` (past-tense) event. The `validate` (present-tense) is used internally to trigger the attached validation cycles for each field.

```js
let validateEvent = new CustomEvent('validate');
for (let field of form.elements) field.dispatchEvent(validateEvent);
```

And each input component can set its own errors based on that.

```js
@tracked validationMessages = [];

*validate(value, field) {
  let { validity: validators = [] } = this.args;
  if (!Array.isArray(validators)) validators = [validators];
  // validator functions return either [] for valid or […] for invalid
  for (let validator of validators) yield* validator(value, field);
  // Capture Vanilla™ constraint messages
  if (field.validationMessage !== '') yield field.validationMessage;
}

@action
onValidateHandler(event, updatedValue) {
  let field = event.target;
  this.validationMessages = Array.from(this.validate(updatedValue, field));
  field.setCustomValidity(isEmpty(this.validationMessages) ? '' : this.validationMessages[0]);
  field.dispatchEvent(new CustomEvent('validated'), { detail: this.validationMesages });
}
```

I hope this helps.

sukima: Was also thinking if you liked the ember-validity-modifier, there isn't any technical reason I know of that you could use it as the validation system for this form project. The concept is detached from the use cases. If you look at the [Demo](https://sukima.github.io/ember-validity-modifier/) all you need is to replace `<form>` with your Ember version of things. Including the custom messaging.

PR #30, opened 2023-02-06
Get ready for automated beta release, enable all C.I. checks

NullVoxPopuli:
Same as: https://github.com/CrowdStrike/ember-toucan-core/pull/14

To exit pre-release mode:
```
pnpm changeset pre exit
```
Docs: https://github.com/changesets/changesets/blob/main/docs/prereleases.md 

Note that not every PR needs a changeset entry, and this is intentional -- especially as we work towards the baseline set of features.

But we do need at least one changeset to have our first beta release.
I've added the Changeset bot which provides instructions on how to make a changeset.

_We do not expect external contributors to provide changesets_.
The best workflow I've seen is for the PR reviewer (if contribution is external) to add the changeset to the PR before merge. 

As a fallback, I've made another tool, demonstrated over here: https://github.com/embroider-build/embroider/pull/1345 
which allows you to double check contributions before release.

---------------------------------

## What's included in this repo's C.I.?

 - assert that source maps and type-maps are generated
 - support multiple versions of typescript, per https://www.semver-ts.org/
 - support multiple versions of ember-source, to ensure a smooth upgrade path for downstream consumers
 - support embroider so that downstream consumers can be assured that their dependencies are "embroider strict mode safe"
- test with no lockfile to try to catch unintended ecosystem deviations / breaking changes (such as what may happen with babel / rollup / etc)

(all of the above is standard in ember OSS projects 🎉 )

extra stuff for this repo:
- publish the docs site and post the preview URLs to PRs
- configure the auto-release system to do a series of beta versions until we're ready for 1.0.0

## What is supported?

- minimum ember-source 4.8 -- we can start pretty new, but this also allows us to use the nicest features. Additionally, since we are beta, we can move the minimum ember-source to whichever ember-source ships the built-in-types (as non-preview), without a breaking change
- TS 4.8 and 4.9 -- we cannot support earlier than TS 4.8, because Glint does not support prior to TS 4.8
- embroider-strict and classic builds

-----------------------------------------

(some of this was already happening -- but this is copy-pasta from toucan-core)

PR #31, opened 2023-02-06
Add `@validateOn="input", rename "blur" to "focusout"

simonihmig:
When playing with validations in a real demo rather than just looking for tests to become green, I realised that the behaviour we likely want for revalidation is not given when using `@revalidationOn="change"`. We want to have revalidation happen as soon as you type (into an invalid field) and it turns valid, _without_ having to focus out of the input (which would effectively be `@revalidationOn="blur"`). But `change` does not trigger right while typing for inputs (normal texts inputs, checkboxes, radios and select boxes this is different though), only when focussing out ("committing"). (see [`change` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event)).

So this introduces another option `@revalidationOn="input"`, listening for the [`input` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event):

image

The PR also takes this as a opportunity to rename `@(re)validationOn="blur"` to `@(re)validationOn="focusout"`, as this is the event we are actually listening to, and having one value for the public API and another for the internal implementation did not really serve a meaningful purpose (other than *maybe* `blur` being a bit more common), but could likely cause confusion.


nicolechung:
Release questions:
1. Are we doing releases yet? Because I guess this would be a breaking change (since we are changing `"blur"` to `"focusout"` but since we haven't released anything yet I guess this is not.
2. If we are doing releases, do we look for the changeset in this PR? (I feel like I have asked this question before in another repo...clearly I don't have a lot of hands on practise with changesets yet)

NullVoxPopuli:
> Are we doing releases yet? 

not yet -- I've opened this: https://github.com/CrowdStrike/ember-headless-form/pull/30
and it starts us out with beta releases, so breaking changes are non-major.

#30 has all the context for how to start releases if we want to publish to npm

test-app/tests/helpers/dom.ts
```diff
@@ -0,0 +1,24 @@
+import { find, triggerEvent } from '@ember/test-helpers';
+
+/**
+ * Fill the provided text into the `value` property of the selected form element, similar to `fillIn`, but *without* implicitly triggering a `change` event.
+ * This mimics the behavior of a user typing data into an input without yet focusing out of it. Browsers will only trigger a `change` event when focusing
+ * out of the element, not while typing!
+ *
+ * `fillIn` will basically simulate entering the data *and* kinda focusing out (as it triggers `change`, however not `focusout`, which is impossible to achieve as a real user),
```
nicolechung, line 8: 😮 I did NOT know this!

test-app/tests/helpers/dom.ts
```diff
@@ -0,0 +1,24 @@
+import { find, triggerEvent } from '@ember/test-helpers';
+
+/**
+ * Fill the provided text into the `value` property of the selected form element, similar to `fillIn`, but *without* implicitly triggering a `change` event.
```
NullVoxPopuli, line 4: 10/10 would read again

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -1128,7 +1130,137 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+              <field.errors data-test-last-name-errors />
+            </form.field>
+            <button type="submit" data-test-submit>Submit</button>
+          </HeadlessForm>
+        </template>);
+
+        assert
+          .dom('[data-test-first-name-errors]')
+          .doesNotExist(
+            'validation errors are not rendered before validation happens on change'
```
nicolechung, line 1249: Suggestion: `on change` -> `on input`. Since this test is now testing for `input` not change.

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -1128,7 +1130,137 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+          );
+        assert
+          .dom('[data-test-last-name-errors]')
+          .doesNotExist(
+            'validation errors are not rendered before validation happens on change'
```
nicolechung, line 1254: Suggestion: `on change` -> `on input`. Since this test is now testing for `input` not change.

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -1128,7 +1130,137 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+          );
+
+        await input('[data-test-first-name]', 'Foo');
+
+        assert
+          .dom('[data-test-first-name-errors]')
+          .exists(
+            { count: 1 },
+            'validation errors appear on focusout when validation fails'
```
nicolechung, line 1263: Suggestion: `on focusout` -> `on input`.

nicolechung: Don't we want to test that there are no errors on the last name field (by accident)?

test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -1622,5 +1754,239 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+          .doesNotExist(
+            'validation errors are not rendered before initial validation happens before form is filled in'
+          );
+        assert
+          .dom('[data-test-last-name-errors]')
+          .doesNotExist(
+            'validation errors are not rendered before initial validation happens before form is filled in'
+          );
+
+        await fillIn('[data-test-first-name]', 'Foo');
```
nicolechung, line 1935: Do we want `input` here or `fillIn`?

nicolechung (approved):
Thanks for catching this!

test-app/tests/helpers/dom.ts
ynotdraw: TIL

test-app/tests/integration/components/headless-form-validation-test.gts
ynotdraw: I think @simonihmig is doing something similar here as the above tests - ensuring `fillIn` _doesn't_ cause errors to render, but below on line 1976 where he's using `input` it does

ynotdraw (commented):
Changes look great to me! Looks like a lingering `@validateOn="blur"` which is causing CI to be red 

```
 error TS2322: Type '"blur"' is not assignable to type 'ValidateOn | undefined'.
```

test-app/tests/integration/components/headless-form-validation-test.gts
nicolechung: I guess it doesn't matter in this case too because the revalidate doesn't start until after `submit`

test-app/tests/helpers/dom.ts
simonihmig: Yeah, had to confirm this for myself by looking up [their source](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/fill-in.ts#L78-L80).

I should probably create an issue. I guess we cannot change the existing behaviour of `fillIn`, as that would break _every_ app. But maybe we can upstream some additional helpers that cover the different real-world use cases better...

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: We do, just a few lines below!

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: Oh ya, thanks! A lot of boilerplate in those tests, that I just copy-pased from the `change` tests. Fixed that!

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: Yeah, it doesn't matter here. And `fillIn` is basically doing more as explained elsewhere (triggering `input` _and_ `change` events), so when even `fillIn` does not make errors show up, then `input` would neither.

simonihmig:
> Changes look great to me! Looks like a lingering @validateOn="blur" which is causing CI to be red

Right, thank you! Fixed that. Good to see that Glint is actually working 🙂

ynotdraw (approved):

test-app/tests/helpers/dom.ts
simonihmig: Filed this: https://github.com/emberjs/ember-test-helpers/issues/1336

Issue #12, opened 2023-01-13
Support for yup-based schema validation


simonihmig:
E.g. through another addon in the monorepo, like `@ember-headless-form/yup` maybe?

NullVoxPopuli:
what all would be needed for yup support?

name-wise, it sounds great, and with `Changesets`, managing releases for multiple things in a monorepo is super easy 🎉

Issue #11, opened 2023-01-13
Support for ember-changeset-validations


simonihmig:
E.g. through another addon in the monorepo, like `@ember-headless-form/changeset` maybe?

PR #32, opened 2023-02-07
Migrate to multi-package layout

simonihmig:
In anticipation of future (imminent!) additions of optional addon-addons like support for `changeset-validations` (#11) or `yup` (#12), this is moving the repo to a multi-package layout, ~renaming the current addon to `@ember-headless-form/core` and~allowing for additional "adapter" packages like e.g. `@ember-headless-form/changeset-validations`.

There are different ways to do this (compare to https://github.com/embroider-build/addon-blueprint/issues/2), this is putting all public packages under `packages/*`, while assuming that we keep just one test-app (we could create a test-app per package also, but I think we can get around that and test all packages in a single app, given how boilerplate-heavy apps are). But if you have other suggestions as to how to structure this, then please share those!

~There is one unresolved thing though, as #30 has config bits like [this](https://github.com/CrowdStrike/ember-headless-form/blob/main/.github/actions/assert-build/action.yml#L37-L41) assuming a single package. So not sure what to do with this. @NullVoxPopuli did we have a similar case before somewhere?~


NullVoxPopuli:
> There is one unresolved thing though, as https://github.com/CrowdStrike/ember-headless-form/pull/30 has config bits like [this](https://github.com/CrowdStrike/ember-headless-form/blob/main/.github/actions/assert-build/action.yml#L37-L41) assuming a single package. So not sure what to do with this. @NullVoxPopuli did we have a similar case before somewhere?

we can modify that to be all the dist directories in the monorepo, I believe. otherwise, we'll want to look at using turborepo [like here](https://github.com/CrowdStrike/ember-headless-table/pull/44), but we _really_ need the dependenciesInjecteed thing resolved before we use turbo.

>  renaming the current addon to @ember-headless-form/core and allowing for additional "adapter" packages like e.g. @ember-headless-form/changeset-validations.

I think we should keep the main addon `ember-headless-form`, it's easier to say and write. I'm 👍 on the adapter names though.

> , but I think we can get around that and test all packages in a single app

sounds good to me! There is still a lot of focus in that one app, so 💯

simonihmig:
> I think we should keep the main addon ember-headless-form, it's easier to say and write. I'm 👍 on the adapter names though.

Yeah, I considered that also briefly, but idk why I dismissed it later, probably for the sake of consistency. But I agree it's better, so I reverted that change now. Off topic: probably should do the same for https://github.com/simonihmig/ember-responsive-image/pull/442

For the CI issue, I now commented out the dedicated build step, and replaced all `actions/download-built-package` steps with just a plain `pnpm build`. I know that's not what we wanted in the first place, but also I don't want to get blocked on this, or spend additional time figuring out a better solution _at this time_. I think the build step is actually pretty fast, so it's not really an issue. Tracking issue: https://github.com/CrowdStrike/ember-headless-form/issues/33

NullVoxPopuli (approved):
>  Off topic: probably should do the same for https://github.com/simonihmig/ember-responsive-image/pull/442

EXCITEMENT!!!! yay

> For the CI issue, I now commented out the dedicated build step, and replaced all actions/download-built-package steps with just a plain pnpm build. I know that's not what we wanted in the first place, but also I don't want to get blocked on this, or spend additional time figuring out a better solution at this time. I think the build step is actually pretty fast, so it's not really an issue. Tracking issue: https://github.com/CrowdStrike/ember-headless-form/issues/33

yeah, all good -- I need to figure out some things, because I'd absolutely love to have turbo manage our cache for us, rather than having to figure out some arcane glob / directory merging silliness.

package.json
```diff
@@ -22,10 +22,6 @@
     "concurrently": "^7.6.0",
     "prettier": "^2.8.3"
   },
-  "workspaces": [
```
nicolechung, line 25: How come this gets removed? I thought it would be "packages/*"

nicolechung:
Question: I guess it's still early stages, but should the package.json name in `packages/ember-headless-form` be updated to any one of:
1. `@ember-headless-form/core`
2. `@ember-headless-form`
3. `@crowdstrike/ember-headless-form`
4. `@crowdstrike/ember-headless-form/core`

nicolechung (approved):
💯  excited about the upcoming packages!

package.json
NullVoxPopuli: this is only used for yarn and npm. pnpm uses the `pnpm-workspaces.yaml` file

nicolechung:
> Question: I guess it's still early stages, but should the package.json name in `packages/ember-headless-form` be updated to any one of:
> 
>     1. `@ember-headless-form/core`
> 
>     2. `@ember-headless-form`
> 
>     3. `@crowdstrike/ember-headless-form`
> 
>     4. `@crowdstrike/ember-headless-form/core`

I guess this part covers that:
>renaming the current addon to @ember-headless-form/core andallowing for additional "adapter" packages like e.g. @ember-headless-form/changeset-validations.

NullVoxPopuli:
we had [already talked about](https://github.com/CrowdStrike/ember-headless-form/pull/32#issuecomment-1421201202) keeping `ember-headless-table` because it sounds nicer, less to type, etc.  🙈

PR #35, opened 2023-02-08
Add opt-in `@dataMode="mutable"`

simonihmig:
Currently the form behaves as a good Ember citizen by strictly following DDAU, i.e. not mutating anything it receives as arguments, especially `@data`. Internally it maintains a copy of that (`internalData`), being passed back to the user only `@onSubmit`.

But there are cases were mutation is needed (or maybe wanted, not so uncommon with forms), especially when that copy-behaviour is provided elsewhere, like with a `changeset` or other buffer/proxy solutions (e.g. https://github.com/yapplabs/ember-buffered-proxy). 

Specifically for #11 we will need to mutate the changeset to make the changeset-validations works, which live on the changeset itself.

### DIFF
--- a/packages/ember-headless-form/src/components/-private/field.ts
+++ b/packages/ember-headless-form/src/components/-private/field.ts
@@ -1,6 +1,6 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
-import { action } from '@ember/object';
+import { action, get } from '@ember/object';
 
 import CaptureEventsModifier from './capture-events';
 import CheckboxComponent from './control/checkbox';
@@ -123,7 +123,8 @@ export default class HeadlessFormFieldComponent<
   }
 
   get value(): DATA[KEY] {
-    return this.args.data[this.args.name];
+    // when @mutableData is set, data is something we don't control, i.e. might require old-school get() to be on the safe side
+    return get(this.args.data, this.args.name) as DATA[KEY];
   }
 
   get errors(): ValidationError<DATA[KEY]>[] | undefined {
--- a/packages/ember-headless-form/src/components/-private/field.ts
+++ b/packages/ember-headless-form/src/components/-private/field.ts
@@ -111,6 +111,11 @@ export default class HeadlessFormFieldComponent<
   ) {
     super(owner, args);
 
+    assert(
+      'Nested property paths in @name are not supported.',
+      typeof this.args.name !== 'string' || !this.args.name.includes('.')
+    );
+
     this.args.registerField(this.args.name, {
       validate: this.args.validate,
     });
@@ -124,6 +129,7 @@ export default class HeadlessFormFieldComponent<
 
   get value(): DATA[KEY] {
     // when @mutableData is set, data is something we don't control, i.e. might require old-school get() to be on the safe side
+    // we do not want to support nested property paths for now though, see the constructor assertion!
     return get(this.args.data, this.args.name) as DATA[KEY];
   }
 
--- a/packages/ember-headless-form/src/components/headless-form.ts
+++ b/packages/ember-headless-form/src/components/headless-form.ts
@@ -2,7 +2,7 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { assert, warn } from '@ember/debug';
 import { on } from '@ember/modifier';
-import { action } from '@ember/object';
+import { action, set } from '@ember/object';
 import { waitFor } from '@ember/test-waiters';
 
 import { modifier } from 'ember-modifier';
@@ -34,6 +34,7 @@ export interface HeadlessFormComponentSignature<DATA extends UserData> {
   Element: HTMLFormElement;
   Args: {
     data?: DATA;
+    mutableData?: boolean;
     validateOn?: ValidateOn;
     revalidateOn?: ValidateOn;
     validate?: FormValidateCallback<FormData<DATA>>;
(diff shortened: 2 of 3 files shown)


packages/ember-headless-form/src/components/headless-form.ts
```diff
@@ -34,6 +34,7 @@ export interface HeadlessFormComponentSignature<DATA extends UserData> {
   Element: HTMLFormElement;
   Args: {
     data?: DATA;
+    mutableData?: boolean;
```
ynotdraw, line 37: Wondering if `mutableData` may be confused with an object rather than a boolean on what it does if folks don't get autocomplete to help them?  Could maybe prefix with `is` or `enable` or switch to like `dataMode?: 'mutable' | 'immutable'`. 

Totally non-blocking, just thinking out loud!

ynotdraw (approved):

packages/ember-headless-form/src/components/headless-form.ts
ynotdraw: Naming is so incredibly hard!  It's the worst 😂. `dataMode` feels a bit better to me but maybe others can chime in later!

packages/ember-headless-form/src/components/-private/field.ts
```diff
@@ -123,7 +123,8 @@ export default class HeadlessFormFieldComponent<
   }
 
   get value(): DATA[KEY] {
-    return this.args.data[this.args.name];
+    // when @mutableData is set, data is something we don't control, i.e. might require old-school get() to be on the safe side
```
nicolechung, line 126: I don't really know the history of the old school ember `get` versus the new getters...is there an article on it?

packages/ember-headless-form/src/components/headless-form.ts
nicolechung: `dataMode` sounds better if other data modes come up (i.e. not just mutable and immutable).

nicolechung (approved):
👍

packages/ember-headless-form/src/components/-private/field.ts
NullVoxPopuli, line 126: > i.e. might require old-school get() to be on the safe side

I'm worried this enables `.`-separated keys, tying us to some form of `get` forever -- could this be something the adapter layer does?

simonihmig: Do you think `get()` will/should go away? Like even in a future with deprecated/removed CPs, it could still exist as a simple util for nested property access? So I am not that worried about that tbh...

Also, I remember we chatted about nested data, which would require just this, not? This is certainly not enough to make nested data officially supported, would require test cases etc, but at the end we would need `get()` here anyway, right?

simonihmig: Oh, you are blessed then! 😅

I don't know of an article that specifically focuses on that. But `get`/`set` (or `this.get()` and `this.set()`) were for a long part essential pieces of Ember before the Octane edition. They were required to access [Computed Properties (CP)](https://api.emberjs.com/ember/release/classes/ComputedProperty), which are still a thing, not yet deprecated, and probably still all around the monorepo codebase, but kinda out-of-favor and superseded by the Octane reactivity model of `@tracked` and native getters. A while ago the need for `get()` to access them was dropped, except for some special cases (in ember-data, idk if that is still required now, I hope not), but `set()` is still needed to set a CP. Also `get()` enabled nested property access, as in `get(myObj, 'some.nested.property')`.

simonihmig: If we don't want to leak the semantics of nested property access here, then we could also add an `assert()` for the path (name) to not contain any dots? 

This way we could at least make nested access an explicit choice later, when we are confident about its use as _our_ public API.

packages/ember-headless-form/src/components/headless-form.ts
simonihmig: I did switch to `@dataMode` now!

packages/ember-headless-form/src/components/-private/field.ts
NullVoxPopuli: >  add an assert() for the path (name) to not contain any dots?

yeah, I'm good with that

> will/should go away?

yeah, I can't speak on behalf of the core team, as I haven't talked with them about this, but -- it's hard to type correctly (and presently has "wrong" types which are "good enough", for performance reasons -- types on nested keys do not say what you'd expect).
In the future, I imagine we'd handle this situation differently, which _is_ kind of like an adapter-like thing.
get does _weird_ and _unexpected_ things with proxying the returned value (iirc) -- it def feels like tech debt when debugging it.

>  it could still exist as a simple util for nested property access? 

for this use case specifically, the core teams have said they want to discourage get usage from ember, and instead defer to other "utility libraries", such as "lodash.get".

>  we chatted about nested data, which would require just this, not?

it would, but we also talked about requiring an adapter and flat data if nested input-data was needed, and I kinda like that explicitness, as native `FormData` can _only_ understand flat data structures (even with dots, the field names are still a single property) -- changeset, and some forms I've built use a notation convention for translating between nested properties / objects / arrays via `.dot.property.0.index.4.access` -- it's.... it required a lot of tests in validated-changeset to get it all working 😅

simonihmig: > add an assert() for the path (name) to not contain any dots?

> yeah, I'm good with that

Did that now!

PR #37, opened 2023-02-10
Add support for changeset-validations

### LINKED ISSUE #11: Support for ember-changeset-validations

simonihmig:
The new `@ember-headless-form/changeset` provides the glue code to integrate the validation capabilities of `ember-changeset` into our forms. Closes #11 

The first commit contains the addon setup, using the new `--addon-only` flag of the v2 addon blueprint courtesy of @NullVoxPopuli, which worked nicely! :tada: This brings in the most boilerplate, so reviewing by commit ignoring the first would probably be easier.


package.json
```diff
@@ -8,10 +8,10 @@
   "scripts": {
     "release": "changeset publish",
     "prepare": "pnpm build",
-    "build": "pnpm --filter ember-headless-form build",
+    "build": "pnpm --filter ember-headless-form build && pnpm --filter '@ember-headless-form/*' build",
```
simonihmig, line 11: This is a cheap workaround for dependency problems, as we need to build the main `ember-headless-form` first before the new one, because the latter consumes types of the former, which would lead to build/type errors when the `/dist` output containing the declaration files is not ready yet. This could be replace by a turborepo based setup in the future...

.changeset/four-squids-change.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ember-headless-form/changeset': major
```
simonihmig, line 2: Out of curiosity I tagged this as a major release, to see what that will lead to. Can change if that doesn't make any sense...

NullVoxPopuli: it does make sense! that'll start us at 1.0.0-beta (since we're in beta mode)

package.json
NullVoxPopuli: does this repo need to do the `pnpm install --force` or `pnpm install` after a build to get around the dependenciesMeta.*.injected issue?

if not, we can set up turbo right away

test-app/tests/integration/components/changeset-test.gts
```diff
@@ -0,0 +1,222 @@
...
+    });
+
+    const data: TestFormData = { firstName: 'Foo', lastName: 'Smith' };
+    const submitHandler = sinon.spy();
+
+    await render(<template>
+      <HeadlessForm
+        @data={{data}}
+        @dataMode="mutable"
+        {{! @glint-expect-error --  a type error is expected here, as this test intentionally has a type mismatch whwithen data not being a changeset }}
```
NullVoxPopuli, line 61:
```suggestion
        {{! @glint-expect-error --  a type error is expected here, as this test intentionally has a type mismatch when data not being a changeset }}
```

packages/changeset-validations/package.json
```diff
@@ -0,0 +1,111 @@
...
+    "test": "echo 'A v2 addon does not have tests, run tests in test-app'",
+    "prepack": "rollup --config"
+  },
+  "dependencies": {
+    "@embroider/addon-shim": "^1.0.0"
+  },
+  "peerDependencies": {
+    "validated-changeset": "^1.3.4",
+    "ember-changeset": "^4.1.2",
+    "ember-headless-form": "^0.0.0"
```
ynotdraw, line 34: This _doesn't_ need to use the `workspace:../` notation as it's a separate package built on top of `ember-headless-form` correct?  (Just trying to wrap my head around that concept a bit more!)

packages/changeset-validations/src/helpers/validate-changeset.ts
```diff
@@ -0,0 +1,51 @@
...
+  fields
+) => {
+  assert(
+    'Cannot use `validateChangeset` on `@data` that is not a Changeset instance!',
+    isChangeset(changeset)
+  );
+
+  // there is also an argument-less version of changeset.validate(), but for this to work the changeset needs a so called validationMap, and not just a validator function
+  // while ember-changeset-validations would provide such a map, we cannot necessarily rely on it being present, so the way to reliably validate all fields is to iterate
+  // over them explicitly
```
ynotdraw, line 18: Thanks for adding this comment - helps a ton!

test-app/tests/integration/components/changeset-test.gts
```diff
@@ -0,0 +1,222 @@
...
+    assert
+      .dom('[data-test-first-name-errors]')
+      .hasText('Foo is an invalid firstName!');
+    assert
+      .dom('[data-test-last-name-errors]')
+      .doesNotExist(
+        'validation errors are not rendered when validation succeeds'
+      );
+  });
+});
```
ynotdraw, line 222: Nice tests!

package.json
simonihmig: I don't use the injected thing at all so far, because... it worked without it 🤷‍♂️

packages/changeset-validations/package.json
simonihmig: I don't think you can use it here, because the `peerDependencies` is a field that users will use when installing the package, so it cannot refer to our monorepo. We have `workspace:*` for `devDependencies` though, which is only used internally, and by which we say that we don't want the package installed from npm under any circumstance, but to always use/link the code that's in the monorepo.

package.json
NullVoxPopuli: legit, I'll have a go at setting up turbo in here soon

packages/changeset-validations/package.json
```diff
@@ -0,0 +1,111 @@
+{
+  "name": "@ember-headless-form/changeset",
```
NullVoxPopuli, line 2: does this package have a README with usage instructions?

simonihmig: No, same as the main one! 😅

I plan to add that as part of the documentation efforts (#4)

packages/changeset-validations/src/helpers/validate-changeset.ts
```diff
@@ -0,0 +1,51 @@
+import { isChangeset } from 'validated-changeset';
```
NullVoxPopuli, line 1: can this file be re-exported from index.ts? makes the public-facing API/imports nicer, imo

packages/changeset-validations/src/template-registry.ts
```diff
@@ -0,0 +1,9 @@
+// Easily allow apps, which are not yet using strict mode templates, to consume your Glint types, by importing this file.
+// Add all your components, helpers and modifiers to the template registry here, so apps don't have to do this.
+// See https://typed-ember.gitbook.io/glint/using-glint/ember/authoring-addons
+
+import type validateChangeset from './helpers/validate-changeset';
+
+export default interface Registry {
+  validateChangeset: typeof validateChangeset;
```
NullVoxPopuli, line 8: because the helper is hyphenated, I think this needs to be hyphenated, too, yeah?

test-app/tests/integration/components/changeset-test.gts
```diff
@@ -0,0 +1,222 @@
...
+
+      if (newValue.toLowerCase() === 'foo') {
+        errors.push(`Foo is an invalid ${key}!`);
+      }
+    }
+
+    return errors.length > 0 ? errors : true;
+  };
+
+  // Could not get this test to work, setupOnerror does not catch the error as expected here
```
NullVoxPopuli, line 43: 🤔 it should 🤔 that's goofy.

test-app/tests/integration/components/changeset-test.gts
```diff
@@ -0,0 +1,222 @@
+/* eslint-disable no-undef -- Until https://github.com/ember-cli/eslint-plugin-ember/issues/1747 is resolved... */
```
NullVoxPopuli, line 1: in a follow up PR, I think it'd be good to test scenarios which use these as well: https://github.com/poteto/ember-changeset-validations

NullVoxPopuli (approved):

test-app/tests/integration/components/changeset-test.gts
NullVoxPopuli: Like, I was expecting to see these:
```
import {
  validatePresence,
  validateLength,
  validateConfirmation,
  validateFormat
} from 'ember-changeset-validations/validators';
```

packages/changeset-validations/src/helpers/validate-changeset.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import type { ErrorRecord, FormValidateCallback } from 'ember-headless-form';
+import type { EmberChangeset } from 'ember-changeset';
+import { assert } from '@ember/debug';
+
+const validateChangeset: FormValidateCallback<EmberChangeset> = async (
```
nicolechung, line 7: Suggestion: Can we get some JSDOC here with a code sample?

Pseudocode, not tested...
```
import validateChangeset from '@ember-headless-form/changeset/helpers/validate-changeset';

const changeset = Changeset(data, validator);

// in template
 <HeadlessForm
        @data={{changeset}}
        @dataMode="mutable"
        @validate={{validateChangeset}}
        @onInvalid={{invalidHandler}}
        as |form|
      >
      ```

File: packages/ember-headless-form/src/components/headless-form.ts [hunk 13]
```diff
@@ -37,7 +37,7 @@ export interface HeadlessFormComponentSignature<DATA extends UserData> {
     dataMode?: 'mutable' | 'immutable';
     validateOn?: ValidateOn;
     revalidateOn?: ValidateOn;
-    validate?: FormValidateCallback<FormData<DATA>>;
+    validate?: FormValidateCallback<DATA>;
```
nicolechung, line 40: Just curious, why did FormData get removed here?

packages/ember-headless-form/src/components/headless-form.ts
```diff
@@ -175,7 +175,10 @@ export default class HeadlessFormComponent<
   @waitFor
   async validate(): Promise<ErrorRecord<FormData<DATA>>> {
     const nativeValidation = this.validateNative();
-    const customFormValidation = await this.args.validate?.(this.internalData);
+    const customFormValidation = await this.args.validate?.(
+      this.internalData,
+      Array.from(this.fields.keys())
```
nicolechung, line 180: This new line is needed for changeset validations?

test-app/tests/integration/components/changeset-test.gts
```diff
@@ -0,0 +1,222 @@
...
+/* eslint-disable simple-import-sort/imports,padding-line-between-statements,decorator-position/decorator-position -- Can't fix these manually, without --fix working in .gts */
+
+import { click, fillIn, render, setupOnerror } from '@ember/test-helpers';
```
nicolechung, line 4: A bit off topic, but I noticed you merged this file for native tests: https://github.com/CrowdStrike/ember-headless-form/pull/29/files#diff-e54bb85c25006594b458dfc882cd3ab24ddaae50f33ab73705b8200ee394709d

But when I look at main this file is now empty??
https://github.com/CrowdStrike/ember-headless-form/blob/main/test-app/tests/integration/components/headless-form-native-validation-test.gts

nicolechung: Asking because when I look here: https://github.com/CrowdStrike/ember-headless-form/actions/runs/4144918649/jobs/7168593225 I don't see the native validation tests

This could be a total Friday brain thing, but I'd feel more confident if I could see that the changeset tests are running beside the native tests.

nicolechung (approved):
Looks great! I left a comment because I can't seem to find the native tests anymore (maybe this is just me).

test-app/tests/integration/components/changeset-test.gts
simonihmig: Uh, wut??? 🤯

Thank you for catching this!!!

This was removed in this commit apparently: https://github.com/CrowdStrike/ember-headless-form/commit/2bbaa709d79b67823611a727241bc7923ca5aa53. The only explanation I have is that I probably rebased the PR shortly before merging, and somehow messed this up badly 🙈

Will add it back in a separate PR!

packages/ember-headless-form/src/components/headless-form.ts
simonihmig: Yeah, tbh, I am not really sure about the types here, they tend to get a bit too complicated maybe... `FormData<>` will limit the keys of the object to only those that are strings. That was needed, because we map the keys to what users specify as their fields `@name`, an the latter can only be strings. But keys of type `object` can besides strings also be `number` or `symbol`. So `FormData` is getting rid of those. 

But `validate` will actually receive the whole `@data` object, including _potentially_ keys of type `number | symbol`, and TS was complaining about mismatching types for the `validateChangset` helper due to this...

Again, not super happy about all of this, as this does not really matter in the real world, it is more to just make TS shut up complaining about subtle type errors 🙄

NullVoxPopuli: what is correct though? We can coerce TS in to doing our bidding (I have lots of experience with this with getting ember-resources and ember-statechart-component glint-compatible 😅 )

packages/ember-headless-form/src/components/headless-form.ts
simonihmig: Yeah, unfortunately. This related to [this comment](https://github.com/CrowdStrike/ember-headless-form/pull/37/files#diff-d0f4e6ca29dc2ef6552d7fe170b4190b35443c13bcca85e0a19f31367bed01a4R16-R20)

Changeset somehow does not know about its own fields, unless you pass a "validationMap". So for this to work in every case, we need to pass the validatable fields. 

Their API is weird, lost too much time to figuring this out. I don't really like their solution tbh. At least for normal forms stuff, this shouldn't be needed. Like our forms already don't mutate the data, but give you the changed data at submit, so you can decide how and when to "commit" those changes. So you don't really need the changeset-behaviors for normal forms usage. And their validation support is also a bit weird, and Ember-centric where validations don't really have to be...

I think our future recommendations for validations should basically be:
* use native as much as possible. For simple forms, that should be even enough
* use `yup` (or `zod`?) for the more advanced stuff (their type inference is lovely!)
* no changesets needed for data "buffering"

packages/changeset-validations/src/helpers/validate-changeset.ts
simonihmig: Good point! Added this ticket: https://github.com/CrowdStrike/ember-headless-form/issues/38

packages/ember-headless-form/src/components/headless-form.ts
NullVoxPopuli: does this work when fields' presence are gated by an if statement?

test-app/tests/integration/components/changeset-test.gts
simonihmig: Yeah, I can do that. However those already use the underlying APIs of ember-changeset that we are using here. So I don't expect the tests to really increase the test coverage, other than testing ember-changeset-validations itself. But maybe it's good to have that nevertheless, for the teaching-aspects of tests (yeah, tests have that! 😀), as they show a more realistic use case?

test-app/tests/integration/components/changeset-test.gts
simonihmig: I have seen errors in tests that could only be catched by plugging into qunit's own error handling IIRC... :think:

packages/changeset-validations/src/template-registry.ts
simonihmig: Right, fixed!

packages/changeset-validations/src/helpers/validate-changeset.ts
simonihmig: Added https://github.com/CrowdStrike/ember-headless-form/issues/39

PR #44, opened 2023-02-13
Setup docs app

simonihmig:
This is only the boilerplate of the docs app, copied from ember-toucan-core, and enabling CI/CD. Writing the actual documentation will follow in separate PR(s).


simonihmig:
@NullVoxPopuli [this message](https://github.com/CrowdStrike/ember-headless-form/actions/runs/4165946401/jobs/7209639446#step:6:13) seems to suggest we need to create a new project for this? Do you have the permissions, I don't have any cloudflare access on behalf of CS, right?

NullVoxPopuli:
> @NullVoxPopuli [this message](https://github.com/CrowdStrike/ember-headless-form/actions/runs/4165946401/jobs/7209639446#step:6:13) seems to suggest we need to create a new project for this? 

I _just_ made it 🎉 

> I don't have any cloudflare access on behalf of CS, right?

Right, we can get you added if you want (UX has its own cloudflare account). But for this kind of setup you _also_ would need to be a GitHub org admin. :(

.github/workflows/ci.yml
```diff
@@ -167,62 +167,49 @@ jobs:
...
+    outputs:
+      url: ${{ steps.publishStep.outputs.url }}
+      env: ${{ steps.publishStep.outputs.environment }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/pnpm
+      - run: pnpm build
+      # - uses: ./.github/actions/download-built-package
```
ynotdraw, line 186: Can this be removed since we have the build `run` above+below?

.github/workflows/ci.yml
```diff
@@ -167,62 +167,49 @@ jobs:
...
+    name: Post Preview URL as comment to PR
+    runs-on: ubuntu-latest
+    needs: PublishDocstoCloudflarePages
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+      - uses: ./.github/actions/pnpm
+      - run: pnpm build
+      # - uses: ./.github/actions/download-built-package
```
ynotdraw, line 208: Same question here as above

.github/workflows/ci.yml
```diff
@@ -167,62 +167,49 @@ jobs:
...
+      - uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          message: |+
+            ## Preview URLs
+            Env: ${{ needs.PublishDocstoCloudflarePages.outputs.env }}
+            Docs: ${{ needs.PublishDocstoCloudflarePages.outputs.url }}
+#            # api docs: ${{ needs.PublishDocstoCloudflarePages.outputs.url }}/api/modules.html
```
ynotdraw, line 215: I should maybe do this as well in toucan-core, as we don't have api docs yet!

docs-app/app/components/feature-card.gts
```diff
@@ -0,0 +1,20 @@
...
+    <h4 class="pb-4 type-2xl text-titles-and-attributes">
+      {{@title}}
+    </h4>
+    <p class="text-body-and-labels">
+      {{yield}}
+    </p>
+  </div>
+</template>;
+
+export default FeatureCard;
```
ynotdraw, line 20: Maybe we can/should move these over to `oss-docs` eventually? (Not part of this PR, obvs, low priority!)

docs-app/app/index.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>ember-headless-form</title>
+    <meta name="description" content="components in ember-toucan-core" />
```
ynotdraw, line 6:
```suggestion
    <meta name="description" content="components in ember-headless-form" />
```

docs-app/ember-cli-build.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+const isProduction = () => EmberApp.env() === 'production';
+
+module.exports = function (defaults) {
+  let app = new EmberApp(defaults, {
+    autoImport: {
+      watchDependencies: ['@crowdstrike/ember-toucan-core'],
```
ynotdraw, line 10:
```suggestion
      watchDependencies: ['ember-headless-form'],
```

docs-app/package.json
```diff
@@ -0,0 +1,129 @@
...
+    "@embroider/router": "^2.0.0",
+    "ember-browser-services": "^4.0.4",
+    "ember-cached-decorator-polyfill": "^1.0.1",
+    "ember-headless-form": "workspace:*",
+    "ember-modifier": "^4.0.0",
+    "ember-resources": "^5.4.0",
+    "highlight.js": "^11.6.0",
+    "highlightjs-glimmer": "^1.4.1",
+    "tracked-built-ins": "^3.1.0"
+  }
```
ynotdraw, line 128: Do you need the `dependenciesMeta` here as well? E.g.,:

```
"dependenciesMeta": {
    "ember-headless-form": {
      "injected": true
    }
  }
```

(https://github.com/CrowdStrike/ember-toucan-core/blob/main/docs-app/package.json#L134)

docs-app/tests/helpers/index.js
```diff
@@ -0,0 +1,42 @@
...
+  // Additional setup for rendering tests can be done here.
+}
+
+function setupTest(hooks, options) {
+  upstreamSetupTest(hooks, options);
+
+  // Additional setup for unit tests can be done here.
+}
+
+export { setupApplicationTest, setupRenderingTest, setupTest };
```
ynotdraw, line 42: Totally non-blocking quesiton: We have all of the same in `toucan-core` under `docs-app/tests`, but I'm curious if it makes sense for our docs-apps to even have the `tests/` directory?  We normally use the `test-app` for testing, so just curious on thoughts around this.

docs-app/.prettierrc.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+module.exports = {
+  plugins: ['prettier-plugin-ember-template-tag'],
```
nicolechung, line 4: !! Should we be using this in our other packages? Does it format everything in between `<template>` tags (is this what it does?)

docs-app/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Installation
+
+- `git clone <repository-url>` this repository
+- `cd docs-app`
+- `npm install`
+
+## Running / Development
+
+- `ember serve`
```
nicolechung, line 23: Question...are we really running `ember serve` or is it `pnpm start`? Or should updating this `README` be in a followup?

docs-app/package.json
nicolechung: It seems like it was working without it based on what I read from a previous PR (let me dig it up)...but then how come we need it in other packages?? (don't feel the need to answer this if it's going to take a lot of investigation btw, just curious)

nicolechung: Update, yeah you are right it was added over here: https://github.com/CrowdStrike/ember-headless-form/pull/49/files#diff-b03f4acb30e04ceb12a40763ef649759b3f632df2357bcec14fab6dd29ebbfb4R124 in the test app

docs-app/tests/helpers/index.js
NullVoxPopuli: over time docs may have demos complex enough where you want them tested

docs-app/README.md
NullVoxPopuli: they are equiv 🎉

docs-app/tests/helpers/index.js
ynotdraw: Ah, that makes sense 👍

.github/workflows/ci.yml
simonihmig: The other jobs also still have this. We should clean up all of this of course, when we have the final solution, which I hope #52 will bring!

docs-app/app/components/feature-card.gts
simonihmig: Yeah, I think this was even mentioned somewhere, in a comment or so. Just vaguely remember...

docs-app/.prettierrc.js
simonihmig: We have this already in the test-app for example. It makes prettier even understand the existence of `<template>`, otherwise it wouldn't be able to parse the file (as it's not valid JS/TS). And yes, it does also format everything inside of the template!

docs-app/README.md
simonihmig: This was copy-paste, and yes, currently there is no difference. However, should we start to add turbo also to scripts in the workspace, including `start`, then it would start to matter!

docs-app/tests/helpers/index.js
simonihmig: Yeah, and I would also like to add even just a very simple smoke test for the docs app, mainly checking that the build does not blow up, and at least something is rendered. This has proven to be useful in other projects like ember-bootstrap, to make sure that e.g. automated renovate updates don't accidentally the docs app!

docs-app/app/index.html
simonihmig: Thanks, applied!

docs-app/package.json
simonihmig: It wasn't needed as I wasn't really using the addon in the docs app yet. But yes, your are right, we should do that here also - and I just did that, to prevent any more foot guns due to this unfortunate issue.

simonihmig:
@NullVoxPopuli I took the freedom to refactor the docs CD workflow a bit, in this [commit](https://github.com/CrowdStrike/ember-headless-form/pull/44/commits/c324e7aa249a8eb6b764bbae37dbc84e92131d37). Previously we had a separate `PostPreviewURLascommenttoPR` job, which seemed too much to me tbh. It did a `pnpm i` and `pnpm build`, which seems unnecessary for the `sticky-pull-request-comment` action. And if you remove those, what is left then is a single step that takes 1s, which IMO doesn't justify a separate job. So I merged the two together.

Btw, that `sticky-pull-request-comment` action wouldn't really be needed, if coudflare would merge and release [this bugfix](https://github.com/cloudflare/pages-action/pull/56), as the cloudflare action will create a `deployment` Github event that makes GH render it as a deployment in its native UI (which due to the bug doesn't work correctly for PRs rn). We could still leave the automated comment in here if we think that's useful though.

ynotdraw (approved):

PR #49, opened 2023-02-15
Add `dependenciesMeta.*.injected`, fix peer deps

simonihmig:
Latest PRs like #45 and #44 are failing CI. After debugging, it seems again related to https://github.com/embroider-build/embroider/issues/1332, in that `dependencySatisfies` (in `ember-modifier` this time) is seeing a wrong package of `ember-source` (that is unused in `test-app`, thus going into that empty package case). 

While a previous attempt to fix this by using `dependenciesMeta.*.injected` before didn't succeed, it seems to do so now!

Also fixing peer deps now, as with `injected` [the workaround of not having ember-source as a devDependency](https://github.com/embroider-build/embroider/issues/1332#issuecomment-1398767613) seems to not be needed anymore.

### DIFF
--- a/packages/changeset-validations/package.json
+++ b/packages/changeset-validations/package.json
@@ -32,7 +32,7 @@
     "validated-changeset": "^1.3.4",
     "ember-changeset": "^4.1.2",
     "ember-headless-form": "^0.0.0",
-    "ember-source": "^4.8.0"
+    "ember-source": "^4.4.0"
   },
   "devDependencies": {
     "@babel/core": "^7.17.0",
--- a/packages/changeset-validations/package.json
+++ b/packages/changeset-validations/package.json
@@ -31,7 +31,8 @@
   "peerDependencies": {
     "validated-changeset": "^1.3.4",
     "ember-changeset": "^4.1.2",
-    "ember-headless-form": "^0.0.0"
+    "ember-headless-form": "^0.0.0",
+    "ember-source": "^4.8.0"
   },
   "devDependencies": {
     "@babel/core": "^7.17.0",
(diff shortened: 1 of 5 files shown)


simonihmig:
I believe CI is failing as we are building the `/dist` output after `pnpm install`, but then the now hard-linked packages [are not synced](https://github.com/pnpm/pnpm/issues/4965#issuecomment-1272634270), so `/dist` is missing. Am I right about this @NullVoxPopuli?

If so, I don't think it's worth it trying to fix this here just to make CI pass, as #45 will refactor the whole setup. But then both PRs are required to make CI green, which means we need to merge one (this?) in a red state and bring the changes into the other to (hopefully!) make things green again. Ok?

packages/changeset-validations/package.json
```diff
@@ -31,7 +31,8 @@
   "peerDependencies": {
     "validated-changeset": "^1.3.4",
     "ember-changeset": "^4.1.2",
-    "ember-headless-form": "^0.0.0"
+    "ember-headless-form": "^0.0.0",
+    "ember-source": "^4.8.0"
```
NullVoxPopuli, line 35: Is the min version 4.4 or 4.8?

The try config has 4.4. last i looked, but if 4.8 is min required, then we can remove the function polyfill 🎉

NullVoxPopuli (approved):

NullVoxPopuli:
> Am I right about this @NullVoxPopuli?

Yup, the download-built-package action, if it's still used, needs to run pnpm i -f again

> Ok

Yup!

ynotdraw (approved):

PR #51, opened 2023-02-15
pnpm i -f must be ran after a build, when using depMea.*.injected

NullVoxPopuli:
due to https://github.com/CrowdStrike/ember-headless-form/pull/49#issuecomment-1431309759, 
https://github.com/CrowdStrike/ember-headless-form/pull/45 is no longer viable.

See: https://github.com/pnpm/pnpm/issues/6088

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/pnpm
-      - run: pnpm build
+      - run: pnpm build && pnpm i -f
       # To be able to run glint, we need the dist directory
       # - uses: ./.github/actions/download-built-package
       - name: Lint + Format + Glint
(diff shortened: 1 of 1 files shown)


ynotdraw (approved):

simonihmig (approved):

simonihmig:
Ok, I could have done that as part of #49. But I thought we would be able to land #45...

This is not feasible currently, as the build step triggered by turbo would need to _also_ have `pnpm i -f`. But still theoretically we could do this? It is just too costly here? More costly than what we have with this PR?

NullVoxPopuli:
> But still theoretically we could do this?

I don't think it's possible -- I opened an issue with turbo here: https://github.com/vercel/turbo/issues/2306 and we tried some solutions, but ultimately, we need to solve: https://github.com/pnpm/pnpm/issues/6088

simonihmig:
Maybe I am misunderstanding this, but couldn't we just modify the `build` script to be something like `"build": "rollup --config && pnpm i -f",`? We are basically doing this here, only at the CI workflow level. Not saying this is nice, but would it work?

NullVoxPopuli:
on a cache-hit, the entirety of the `build` script would be skipped, and the links in `node_modules/.pnpm` not updated

simonihmig:
Oh, right. This was the missing bit 😔

PR #52, opened 2023-02-16
Setup turborepo, second try

simonihmig:
Another attempt, using a custom sync step to work around https://github.com/vercel/turbo/issues/2306

* builds on top of #45 
* adds a `sync-pnpm` tool to sync the `/dist` output
* adds an intermediate turbo task `_build` to call sync after `build`

**Best reviewed by commit** (first one is #45 squashed into one)


packages/changeset-validations/package.json
```diff
@@ -101,7 +101,8 @@
       "types": "./dist/*.d.ts",
       "default": "./dist/*.js"
     },
-    "./addon-main.js": "./addon-main.cjs"
+    "./addon-main.js": "./addon-main.cjs",
+    "./package.json": "./package.json"
```
simonihmig, line 105: This is a workaround for an unfortunate issue...

`sync-pnpm` needs to find the hard linked package resolvable from the test-app, which is like a copy of our original addon in the monorepo, just without `/dist` (which is the issue we need to fix here). But `require.resolve('ember-headless-form')` does not work, because its `main` field points to its `./dist/index.js`, which des not exist before we have actually synced. So `resolve` throws. 

Then I tried with this package that I found on npm, which basically tries to solve that issue: [resolve-pkg](https://github.com/sindresorhus/resolve-pkg). But running into https://github.com/sindresorhus/resolve-pkg/issues/9.

So the workaround here is to make `mypackage/package.json` resolvable. Not ideal, but works for now. Not sure if we can do this in any other way. We want to delegate the node resolution algorithm to node itself and not re-invent this, but there seems to be no way to say "give me the module path that you _would_ try to require, without actually throwing when it is not available".

NullVoxPopuli: does this one support exports? https://www.npmjs.com/package/resolve-package-path -- it's what auto-import uses

NullVoxPopuli:
Total runtime is 22m: https://github.com/CrowdStrike/ember-headless-form/actions/runs/4194427919/usage

can you push an empty commit? hopefully that'd be a full cache hit

packages/changeset-validations/package.json
simonihmig: It seems to replicate the node module resolution, instead of delegating to node (like `require.resolve`), so _probably_ that would work! Thanks for the suggestion, will try this out!

Related: https://github.com/sindresorhus/resolve-pkg/issues/9#issuecomment-1433250154

simonihmig: > will try this out!
 
It works! :tada:

simonihmig:
> Total runtime is 22m:

For the record, caching seems to be fixed now. Total runtime < 9min - https://github.com/CrowdStrike/ember-headless-form/actions/runs/4196887915/usage

NullVoxPopuli (approved):
Nice work!

Would be cool to see on ember-headless-table and ember-toucan-core, as well 🎉

simonihmig:
> Would be cool to see on ember-headless-table and ember-toucan-core, as well 🎉

Yeah, which brings us to the topic of what to do with `sync-pnpm`. I put it here as a PoC. But if that's the thing we use to fix the issue, then we need to extract that to _somewhere_!?

NullVoxPopuli:
>  then we need to extract that to somewhere!?

I wonder if the turbo-repo folks would be interested in adopting it. I'll ask here: https://github.com/vercel/turbo/issues/2306

simonihmig:
I'll hold this back until we have #44 merged, so I can integrate the docs setup here too...

internal/sync-pnpm/index.js
```diff
@@ -0,0 +1,41 @@
+import { join, dirname } from 'node:path';
+import { createRequire } from 'node:module';
+
+import { getPackages } from '@manypkg/get-packages';
+import { findRoot } from '@manypkg/find-root';
+import { readJson, pathExists } from 'fs-extra/esm';
+import { hardLinkDir } from '@pnpm/fs.hard-link-dir';
```
ynotdraw, line 7: Ahhhh - this is interesting!

> Hard link (or copy if linking fails) all files from a directory to several target directories.

So essentially this sync file forces `pnpm` to hard link the dependencies one way or another so that we don't have to continually `pnpm i -f` anytime we make a change in the addon source and want to see it in the test-app/docs-app?

ynotdraw (approved):

internal/sync-pnpm/index.js
simonihmig: Yes, exactly. Also it does _only_ this, it does not try to re-install/shuffle things around. 

In an earlier attempt, I tried to just use `pnpm i -f` in a turbo step, but when things run in parallel, this would lead to weird errors, as `pnpm i -f` would temporarily remove things from `node_modules` and add them back moments later, but in the meantime some parallel build would just fail as it would miss some dependencies in `node_modules`.

So this does just adds back the missing hard links of our `/dist` folder, and nothing else.

nicolechung: Amazing!!

So is the current development process now (for something like `docs-app`):
1. make a change in `ember-headless-form`
2. `pnpm build`
3. see the update in the `docs-app`?

nicolechung (approved):

internal/sync-pnpm/index.js
simonihmig: Yes, exactly, that should work (assuming you have `pnpm start` or `pnpm start:docs` running, to continuously rebuild the app).

It is a bit of a regression compared to other v2 addon setups (e.g. using `yarn`), where this `injected` thing does not exist (but instead other peer dependency related problems), so you could just run `start` and it would watch the addon and the test-app. But not a regression compared to the previous state of this addon, because we had to run `pnpm i -f` again and again, so this should make it better at least!

simonihmig: Just realized that the `pnpm build` script does too much for this use case, as it builds the addons _and_ the apps (test and docs). Will need another iteration!

PR #53, opened 2023-02-17
Add JS/TSDoc comments

### LINKED ISSUE #38: Add better JSDoc documentation

simonihmig:
Gives us nice IntelliSense tooltips in VS Code with Glint:

image

Also enables us to use TypeDoc later for API docs. 

Tried to add https://github.com/hosseinmd/prettier-plugin-jsdoc, but it seems to cause more problems than it solves (messes up code blocks in markdown, also other bugs on its Github issues page)

Closes #38

### DIFF
--- a/packages/changeset-validations/src/helpers/validate-changeset.ts
+++ b/packages/changeset-validations/src/helpers/validate-changeset.ts
@@ -8,7 +8,7 @@ import { assert } from '@ember/debug';
  * Validation helper for integrating `ember-changeset` based validations into headless forms:
  *
  * - pass a changeset to the form's `@data`
- * - pass this helper into the form's `@validate` hook
+ * - pass this helper into the form's `@validate` hook `@validate={{validateChangeset}}`
  * - opt-in to `@dataMode="mutable"`
  */
 const validateChangeset: FormValidateCallback<EmberChangeset> = async (
--- a/packages/changeset-validations/src/helpers/validate-changeset.ts
+++ b/packages/changeset-validations/src/helpers/validate-changeset.ts
@@ -4,6 +4,13 @@ import type { ErrorRecord, FormValidateCallback } from 'ember-headless-form';
 import type { EmberChangeset } from 'ember-changeset';
 import { assert } from '@ember/debug';
 
+/**
+ * Validation helper for integrating `ember-changeset` based validations into headless forms:
+ *
+ * - pass a changeset to the form's `@data`
+ * - pass this helper into the form's `@validate` hook
+ * - opt-in to `@dataMode="mutable"`
+ */
 const validateChangeset: FormValidateCallback<EmberChangeset> = async (
   changeset,
   fields
--- a/packages/ember-headless-form/src/-private/components/control/input.ts
+++ b/packages/ember-headless-form/src/-private/components/control/input.ts
@@ -3,7 +3,7 @@ import { assert } from '@ember/debug';
 import { action } from '@ember/object';
 
 // Possible values for the input type, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types
-// for the sake ok completeness, we list all here, with some commented out that are better handled elsewhere, or not at all...
+// for the sake of completeness, we list all here, with some commented out that are better handled elsewhere, or not at all...
 export type InputType =
   // | 'button' - not useful as a control component
   // | 'checkbox' - handled separately, for handling `checked` correctly and operating with true boolean values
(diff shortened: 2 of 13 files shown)


packages/ember-headless-form/src/template-registry.ts
```diff
@@ -5,5 +5,35 @@
 import type HeadlessFormComponent from './components/headless-form';
 
 export default interface Registry {
+  /**
+   * Headless form component.
```
simonihmig, line 9: Seems for this to appear in a template when hovering `<HeadlessForm>`, we need to place it here, instead of in front of the class definition (where we have it also).

packages/ember-headless-form/src/-private/components/control/input.ts
```diff
@@ -31,12 +31,54 @@ export type InputType =
...
+     * - `radio`
+     *
+     * Also these types are not useful to use as input controls:
+     * - `button`
+     * - `file`
+     * - `image`
+     * - `reset`
+     * - `submit`
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types
```
ynotdraw, line 48: I like the `@see` for docs - nice, TIL

packages/ember-headless-form/src/-private/components/field.ts
```diff
@@ -36,47 +36,159 @@ export interface HeadlessFormFieldComponentSignature<
...
+         * The current value of the field's form data.
+         *
+         * If you don't use one of the supplied control components, then use this to pass the value to your custom component.
+         */
         value: DATA[KEY];
+
+        /**
+         * Action to update the (internal) form data for this field.
+         *
+         * If you don't use one of the supplied control components, then use this to update the value whenever you custom component's value has changed.
```
ynotdraw, line 145:
```suggestion
         * If you don't use one of the supplied control components, then use this to update the value whenever your custom component's value has changed.
```

packages/ember-headless-form/src/template-registry.ts
ynotdraw: (Thank you! I'll remember this for toucan-core stuff! 🎉  )

packages/changeset-validations/src/helpers/validate-changeset.ts
```diff
@@ -4,6 +4,13 @@ import type { ErrorRecord, FormValidateCallback } from 'ember-headless-form';
 import type { EmberChangeset } from 'ember-changeset';
 import { assert } from '@ember/debug';
 
+/**
+ * Validation helper for integrating `ember-changeset` based validations into headless forms:
+ *
+ * - pass a changeset to the form's `@data`
+ * - pass this helper into the form's `@validate` hook
```
ynotdraw, line 11: Super nit, no need to change

Could this sentence be updated to align closer with the `yup` integration one to show an example of this:

```suggestion
 * - pass this helper into the form's `@validate` hook `@validate={{validateChangeset}}`
```

ynotdraw (approved):
Very well written, thanks for adding these.  Having the hints right inside VSCode is great for DX! 🎉 🚀

packages/ember-headless-form/src/components/headless-form.ts
```diff
@@ -85,6 +121,36 @@ class FieldData<
   validate?: FieldValidateCallback<DATA, KEY>;
 }
 
+/**
+ * Headless form component.
+ *
+ * @example
+ * Usage example:
+ *
+ * ```hbs
```
nicolechung, line 130: 🎉  This is great!

packages/ember-headless-form/src/template-registry.ts
nicolechung: I did not know this either, thank you for pointing it out!

nicolechung (approved):

packages/ember-headless-form/src/-private/components/field.ts
```diff
@@ -36,47 +36,159 @@ export interface HeadlessFormFieldComponentSignature<
...
+         */
         setValue: (value: DATA[KEY]) => void;
+
+        /**
+         * Unique ID of this field, used to associate the control with its label.
+         *
+         * If you don't use the supplied components, then you can use this as the `id` of the control and the `for` attribute of the `<label>`.
```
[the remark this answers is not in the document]
simonihmig: "Control" is not really the tag here, it's more the overarching term we chose for the form element the user will change the value with, like `<input>`, `<select>` or `<textarea>`.

simonihmig (commented):
Thanks for the suggestions, I just applied most of them!

PR #60, opened 2023-02-27
Yield async validation and submission state

simonihmig:
Validation and submission can both be async. This yields the async state (pending/resolved) back to the template, so you can react to it, like disabling a submit button or changing labels.

This uses https://github.com/tracked-tools/ember-async-data, and exposes their interface as our own public API, i.e. users will directly be able to use e.g.`.isPending`. This means we are not hiding it as an implementation detail, so we also cannot remove it without breaking things. But I think this is acceptable, as we also don't want to reinvent things, especially not its very good [TypeScript support](https://github.com/tracked-tools/ember-async-data#with-typescript). Therefore the tests also don't cover all the details of _their_ API, as they already have good test coverage.

To Do:

- [x]  add docs after #58 is merged

### DIFF
--- a/docs/usage/async/demo/async.md
+++ b/docs/usage/async/demo/async.md
@@ -1,0 +1,53 @@
+# Async state
+
+Submit this form with a valid email, and with the same email again, to see how it disables the submit button, changes its label, and shows error messages coming from the "backend":
+
+```hbs template
+<HeadlessForm @onSubmit={{this.handleSubmit}} as |form|>
+  <form.Field @name='email' as |field|>
+    <div class='my-2 flex flex-col'>
+      <field.Label>Email</field.Label>
+      <field.Input
+        @type='email'
+        placeholder='Please enter your email'
+        class='border rounded px-2'
+      />
+    </div>
+  </form.Field>
+
+  <button type='submit' disabled={{form.submissionState.isPending}}>
+    {{if form.submissionState.isPending 'Submitting...' 'Submit'}}
+  </button>
+
+  {{#if form.submissionState.isResolved}}
+    <p>We got your data! 🎉</p>
+  {{else if form.submissionState.isRejected}}
+    <p>⛔️ {{form.submissionState.error}}</p>
+  {{/if}}
+</HeadlessForm>
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class MyFormComponent extends Component {
+  saved = [];
+
+  @action
+  async handleSubmit({ email }) {
+    // pretending something async is happening here
+    await new Promise((r) => setTimeout(r, 3000));
+
+    if (!email) {
+      throw new Error('No email given');
+    }
+
+    if (this.saved.includes(email)) {
+      throw new Error(`${email} is already taken!`);
+    }
+
+    this.saved.push(email);
+  }
+}
+```
--- a/docs/usage/async/demo/async.md
+++ b/docs/usage/async/demo/async.md
@@ -44,6 +44,7 @@ export default class MyFormComponent extends Component {
     }
 
     if (this.saved.includes(email)) {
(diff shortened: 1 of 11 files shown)


simonihmig:
There is still an issue with an embroider scenario failing (seems more like an upstream bug), which I need to figure out (or work around). But this can already be reviewed nevertheless!

patches/ember-auto-import@2.6.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/js/webpack.js b/js/webpack.js
+index 86641517cc55be2349fbed507ec6f93ebd027469..06ba89d05b908b44aa97a7ea6e4aa5a8d0e3101c 100644
+--- a/js/webpack.js
++++ b/js/webpack.js
+@@ -384,7 +384,7 @@ class WebpackBundler extends broccoli_plugin_1.default {
+          */
+         let host = this.opts.rootPackage;
+         let emberSource = host.requestedRange('ember-source');
+-        let emberSourceVersion = semver_1.default.coerce(emberSource);
++        let emberSourceVersion = semver_1.default.valid(emberSource);
```
simonihmig, line 10: This patch can be removed once https://github.com/ef4/ember-auto-import/pull/569 lands.

test-app/package.json
```diff
@@ -74,6 +74,7 @@
     "broccoli-asset-rev": "^3.0.0",
     "concurrently": "^7.6.0",
     "ember-a11y-testing": "^5.2.0",
+    "ember-async-data": "^0.7.0",
```
simonihmig, line 77: Adding this should not be necessary, but it fixes a weird issue making Embroider with _safe_ settings not fail the build, see https://github.com/embroider-build/embroider/issues/1370

test-app/tests/integration/components/headless-form-async-test.gts
```diff
@@ -0,0 +1,329 @@
...
+            {{#if form.validationState.isResolved}}
+              <div data-test-validation-value>
+                {{stringify form.validationState.value}}
+              </div>
+            {{/if}}
+          {{/if}}
+        </HeadlessForm>
+      </template>);
+
+      assert
```
nicolechung, line 88: ✅

docs/usage/async/demo/async.md
```diff
@@ -0,0 +1,53 @@
...
+  async handleSubmit({ email }) {
+    // pretending something async is happening here
+    await new Promise((r) => setTimeout(r, 3000));
+
+    if (!email) {
+      throw new Error('No email given');
+    }
+
+    if (this.saved.includes(email)) {
+      throw new Error(`${email} is already taken!`);
```
nicolechung, line 47: Suggestion, total nit: add a comment like, "Throwing this error will cause the form to yield a `form.validationState. isRejected` and `form.submissionState.isRejected`.

Reason being is I had to read this code a few times to realize that's what this is doing.

docs/usage/async/demo/async.md
```diff
@@ -0,0 +1,53 @@
...
+  {{/if}}
+</HeadlessForm>
+```
+
+```js component
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class MyFormComponent extends Component {
+  saved = [];
```
nicolechung, line 35: Total nit, suggestion: I think it might be worth explaining a bit here about how this is not `@tracked` (since it's not used directly in the template) but that it affects the forms internal state (which is tracked)

nicolechung (approved):
💯  the demo works great! Thank you for adding this.

I left a few suggestions, mostly over stuff that confused me a little until I read the code a few times.

docs/usage/async/demo/async.md
simonihmig: Thanks, applied your suggestion!

PR #61, opened 2023-02-27
Fix sync-pnpm when already synced previously

simonihmig:
The sync script would skip syncing the `dist` folder when it already exists. This is not good when running things locally, like when you run the test-app, and build the addon in watch mode, you would still need to run the sync script after changes in the addon manually. But without this fix here, it would not update the `dist` output with the addon changes.

Currently my development workflow with this is:
* `pnpm start` (runs test-app)
* in another terminal:
  * `cd <addon-dir>`
  * `pnpm start` (rollup in watch mode)
 * in yet another terminal:
   * `cd test-app`
   * on every change of the addon run manually: `pnpm _syncPnpm`

Far from good, but at least works...

### DIFF
--- a/internal/sync-pnpm/index.js
+++ b/internal/sync-pnpm/index.js
@@ -3,11 +3,13 @@ import { createRequire } from 'node:module';
 
 import { getPackages } from '@manypkg/get-packages';
 import { findRoot } from '@manypkg/find-root';
-import { readJson, pathExists } from 'fs-extra/esm';
+import { readJson, pathExists, remove } from 'fs-extra/esm';
 import { hardLinkDir } from '@pnpm/fs.hard-link-dir';
 import resolvePackagePath from 'resolve-package-path';
+import Debug from 'debug';
 
 const require = createRequire(import.meta.url);
+const debug = Debug('sync-pnpm');
 
 const syncDir = './dist';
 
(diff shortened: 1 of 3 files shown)


nicolechung:
Suggestion (maybe I should file an issue for this): Can we change the port of either the docs-app or the test-app so I can run both at the same time?

nicolechung:
> on every change of the addon run manually: pnpm _syncPnpm

Wait...what is the difference between running `pnpm i -f` and `pnpm_syncPnpm` between each build of the package?

ynotdraw (approved):
Nice! I also found success with using `pnpm link`, but not sure if it's any better DX wise.  Here's what I found with ember-toucan-core:

```bash
# add-on
cd ember-toucan-core
pnpm link .

# go to docs-app now and link it
cd ../docs-app
pnpm link @crowdstrike/ember-toucan-core
```

Now build the add-on in watch mode

```bash
cd ember-toucan-core
pnpm start
```

Then in another terminal, start the docs app:

```bash
cd docs-app
pnpm start
```

Now if you make a change in `ember-toucan-core`, it auto-rebuilds and the updates are shown in the `docs-app`!

---

FWIW, I just tried this in a super small example, not with actually building anything real, so there may still be issues with `link`

nicolechung:
Screen Shot 2023-02-27 at 12 04 47 PM

I know that I should know this given I reviewed the earlier version, but how do I run `syncPnpm`

Attached is an image of what I have tried

nicolechung:
> Screen Shot 2023-02-27 at 12 04 47 PM
> 
> I know that I should know this given I reviewed the earlier version, but how do I run `syncPnpm`
> 
> Attached is an image of what I have tried

NVM Tony just showed me its in the app directory...

simonihmig:
> NVM Tony just showed me its in the app directory...

Yeah, it's only available to apps. Ideally we shouldn't call this ourselves of course (hence the underscored name), but we are still figuring things out. Seems @NullVoxPopuli is also up for something:
* https://github.com/CrowdStrike/ember-headless-table/pull/145
* https://github.com/pnpm/pnpm/pull/6135

nicolechung (approved):
Okay now that I figured out how to get this running...this is definitely an improvement!

I only have to run the command `pnpm _synchPnpm` once and it works!

PR #62, opened 2023-02-28
Guard sync-pnpm using lockfile

simonihmig:
When `sync-pnpm` runs in parallel for two apps (only surfaced on #58), there seem to be race conditions possible where [creating hard links fails with an `EEXIST: file already exists` error](https://github.com/CrowdStrike/ember-headless-form/actions/runs/4281841214/jobs/7456567249#step:6:66). This adds a lockfile-based guard so two sync-pnpm cannot run on the same target folder in parallel. 

Alternatively that could maybe also be fixed upstream at `@pnpm/fs.hard-link-dir` (they are already ignoring a couple of error codes specifically), but maybe this is supposed to be like it is for pnpm's own use cases, so going ahead with this fix for now...

/cc @NullVoxPopuli

### DIFF
--- a/internal/sync-pnpm/index.js
+++ b/internal/sync-pnpm/index.js
@@ -7,6 +7,7 @@ import { readJson, pathExists, remove } from 'fs-extra/esm';
 import { hardLinkDir } from '@pnpm/fs.hard-link-dir';
 import resolvePackagePath from 'resolve-package-path';
 import Debug from 'debug';
+import lockfile from 'proper-lockfile';
 
 const require = createRequire(import.meta.url);
 const debug = Debug('sync-pnpm');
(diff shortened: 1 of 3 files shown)


internal/sync-pnpm/index.js
```diff
@@ -37,13 +38,25 @@ export default async function syncPnpm(dir = process.cwd()) {
...
     if (await pathExists(syncFrom)) {
+      let releaseLock;
+      try {
+        releaseLock = await lockfile.lock(syncTo, { realpath: false });
+        debug(`lockfile created for syncing to ${syncTo}`);
+      } catch (e) {
+        debug(
+          `lockfile already exists for syncing to ${syncTo}, some other sync process is already handling this directory, so skipping...`
+        );
+        continue;
```
nicolechung, line 49: 🤯  I didn't even know you could `continue` in a catch statement.

nicolechung (approved):

PR #69, opened 2023-03-02
Rename packages to be under `@crowdstrike/*` scope

simonihmig:
Some clever guy snatched the `@ember-headless-form/*` scope on npm, so as was recommended to me we are going the safe way and put everything under `@crowdstrike/*`.

This includes `@crowdstrike/ember-headless-form` instead of `ember-headless-form`, to make things consistent. There was allready `0.0.0` of `ember-headless-form` published (https://www.npmjs.com/package/ember-headless-form), which we should probably try to get unpublished then.


NullVoxPopuli:
if `ember-headless-form` is to be used outside of the CrowdStrike org, I don't think it makes sense to also prefix it. :-\

simonihmig:
> I think it would undermine the ecosystem-wide efforts/impact to provide a good forms implementation via this package by releasing under any company npm scope

That's certainly not want we want. I haven't perceived it really as such a blocker. But you are right, this is kinda unusual in the ecosystem, at least if the package's concern is not directly related to the company.

Originally the scope was used to make the adapter packages names shorter. But with names like `@crowdstrike/ember-headless-form-yup` there is not really a benefit from the scope, we could instead release just as `ember-headless-form-yup` I guess?

NullVoxPopuli:
> just as ember-headless-form-yup I guess?

yeah, this is a good compromise, and aligns nicely with existing ecosystem patterns, like ember-service-worker* and ember-changeset*

PR #70, opened 2023-03-03
Rename packages and prepare for release

simonihmig:
This removes the `@ember-headless-form/*` package scope for known reasons and adds missing Readme's for the three packages. 

The main package has its Readme at the root level, so you see it immediately on Github. The rollup build copies it into the workspace, so the package published to npm has it also, to be viewed on https://www.npmjs.com/package/ember-headless-form (this setup has always been in place as part of the addon blueprint). The additional adapter packages have their own Readme directly in the workspace folder.


README.md
```diff
@@ -1,10 +1,23 @@
 # ember-headless-form
 
-[Short description of the addon.]
+[![CI](https://github.com/CrowdStrike/ember-headless-form/actions/workflows/ci.yml/badge.svg?branch=main&event=push)](https://github.com/CrowdStrike/ember-headless-form/actions/workflows/ci.yml)
+
+The Ember.js addon that distills the common behavior and accessibility best practices of forms into reusable components, without any opinions on specific markup or styling. Use it to build your forms directly, or to build your opinionated forms component kit on top of it.
```
nicolechung, line 5: This might confuse things, but I wonder if we should explicitly tell folks to not use the @ember-headless-form prefix?

I don't know which is why I am asking...

ynotdraw (approved):

nicolechung (approved):
:tada !

PR #76, opened 2023-03-08
Improve a11y of radio group

### LINKED ISSUE #22: Improve a11y for radio group
#16 added support for radio buttons, with plain markup using `<input type="radio">` and `<label>` (for each radio button), but nothing more. The validation support of #21 was supposed to add proper ARIA attributes for validation and error messages (`aria-invalid`, `aria-describedby`), but it turns out we need to further investigate how to provide accessible markup for radio groups:
* should we enforce having either a `<fieldset>` and `<legend>` to group the radio buttons, _or_ plain divs (might be better for a headless addon, as they don't come with default styling!?) with `role=radiogroup` and `aria-labbeledby` (label is also visible to sighted users) / `aria-label` (sr-only)
* where to attach `aria-describedby` to? All radio buttons, or to `role=radiogroup`?
* similarly, where to attach `aria-invalid` to? All radio buttons, or to `role=radiogroup`?

Related resources:
* https://www.w3.org/WAI/ARIA/apg/patterns/radiobutton/
* https://adrianroselli.com/2022/02/support-for-marking-radio-buttons-required-invalid.html
* https://blog.tenon.io/accessible-validation-of-checkbox-and-radiobutton-groups/
* 🔐 https://docs.google.com/document/d/1mVhx7WDL1-ZM8rDnSsFSEwt_DZ0lxyUtxmQZPbPLQ8Y/edit#heading=h.21k3ok5d8ffj

simonihmig:
Fixes #22


nicolechung:
Something to think about, maybe as a follow up?

https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role#keyboard_interactions

I didn't try to manually test keyboard to confirm it works here...

nicolechung:
Actually...despite what I mentioned earlier, I just checked out the test-app and it looks like keyboard works as expected.

keyboard

Would be nice to get a follow up PR for the tests of it.

nicolechung (approved):
🎉

simonihmig:
> I didn't try to manually test keyboard to confirm it works here...

Yeah, so I think there are quite a number of a11y/ARIA examples out there that are quite misleading, in that they show how things _can_ be done with a ton of ARIA attributes and JS, to make things accessible _despite_ not using the "platform features". Like in this MDN doc you also linked to, with an [example](https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role#examples) that uses only divs and spans to implement radio buttons. 

In the following examples there you have...

> This could have been written using semantic HTML, which requires no CSS or JavaScript:

And I think that's what we have to do (and are doing): use semantic HTML as much as possible, and only add sprinkles of ARIA attributes on top where needed and where it improves things (like what we do with e.g. `aria-invalid`, `aria-describedby` etc.).

So long story short, I don't think we have a problem with keyboard navigation (and as you have confirmed, thanks for double checking!), as we are using "the platform" (like `<input type="radio">`) instead of meaningless divs, and the browser should make that work ootb.

@noahgelmancs you agree here?

> Would be nice to get a follow up PR for the tests of it.

Well, two things here:
1. idk how much we really want to test things that are essentially browser features (see above), taking the benefit/cost ratio into account
2. IIRC, it was not working to programmatically test a user moving focus to the next interactive element by pressing the tab key. Like you could do an `await triggerKeyEvent(...)` or something like this in a test, and the event would actually be there and bubble up the DOM etc, but I think it would _not_ make the browser actually move the focus. So it treats that programmatically created event differently than a user initiated tab press. Which makes testing this pretty much impossible. Again, AFAIK and IIRC... 😁

ynotdraw (approved):

PR #77, opened 2023-03-08
Yield `rawErrors` for custom error rendering

simonihmig:
This closes the gap we found today when trying to wire up error rendering with ember-toucan-core. Instead of using the `<field.Errors>` component, we can now use `field.rawErrors` to get access to the array of raw `ValidationError` instances.

### DIFF
--- a/test-app/tests/integration/components/headless-form-validation-test.gts
+++ b/test-app/tests/integration/components/headless-form-validation-test.gts
@@ -2030,6 +2030,106 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
       assert.dom('[data-test-invalid]').doesNotExist();
     });
 
+    test('form yields rawErrors', async function (assert) {
+      const data: TestFormData = { firstName: 'foo' };
+
+      await render(<template>
+        <HeadlessForm
+          @data={{data}}
+          @validate={{validateFormCallbackSync}}
+          as |form|
+        >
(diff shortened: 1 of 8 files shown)


test-app/tests/integration/components/headless-form-validation-test.gts
```diff
@@ -2030,6 +2030,106 @@ module('Integration Component HeadlessForm > Validation', function (hooks) {
...
+    test('form yields rawErrors', async function (assert) {
+      const data: TestFormData = { firstName: 'foo' };
+
+      await render(<template>
+        <HeadlessForm
+          @data={{data}}
+          @validate={{validateFormCallbackSync}}
+          as |form|
+        >
+          {{#each-in form.rawErrors as |name errors|}}
```
nicolechung, line 2042: 👀 ! I didn't know about `{{each-in}}`. Looks useful!

nicolechung (approved):

test-app/tests/integration/components/headless-form-validation-test.gts
simonihmig: Yeah, so it _might_ actually be the first time I used it in my idk ~10y of working w/ Ember! 😅. (and maybe I just forgot about it, but you definitely don't see very often)

ynotdraw (approved):
🚀

PR #80, opened 2023-03-13
Add `@ignoreNativeValidation` for opting out of native validation

simonihmig:
This should fix the other issue we identified when integration headless-form and toucan-core.


packages/ember-headless-form/src/components/headless-form.ts
```diff
@@ -62,6 +62,13 @@ export interface HeadlessFormComponentSignature<
      */
     validate?: FormValidateCallback<DATA>;
 
+    /**
+     * Allows you to opt-out of native validation.
+     *
+     * This can be useful if all of the validation logic is already handled by the `@validate` hooks, but you have form controls that have validation requirements (e.g. `email` type) that would cause the native validation to interfere or.
```
ynotdraw, line 68: Lingering "or"?

```suggestion
     * This can be useful if all of the validation logic is already handled by the `@validate` hooks, but you have form controls that have validation requirements (e.g. `email` type) that would cause the native validation to interfere.
```

ynotdraw (approved):

docs/validation/native.md
```diff
@@ -33,3 +33,7 @@ In the following example, we use an `email` typed input and applied `required` a
 ```
 
 > Note that the validation messages are generated by the browser, so they are dependent on the browser itself **and** the operating system's locale. Unfortunately there is no easy way to customize them or to force a specific locale (like the page's locale instead).
+
+In case you have also provided [custom validations](./custom-validation.md) via the `@validate` argument, the validation errors from all sources including the native ones will be merged together. So you can have basic validations like setting `required` using native validations and add additional ones on top that are not possible otherwise using custom validation functions or the related adapter packages like [ember-changeset](./ember-changeset.md) or [yup](yup.md).
```
nicolechung, line 37: Suggestion: before line 37, add a header

```
## ignoring native validation
```

test-app/tests/integration/components/headless-form-native-validation-test.gts
```diff
@@ -472,6 +472,81 @@ module(
       assert.dom('[data-test-last-name-errors]').doesNotExist();
     });
 
+    test('opt out of native validation', async function (assert) {
```
nicolechung, line 475: Suggestion: `'opt out of native validation'` -> `'opt out of native validation using @ignoreNativeValidation'`

packages/ember-headless-form/src/components/headless-form.ts
```diff
@@ -62,6 +62,13 @@ export interface HeadlessFormComponentSignature<
...
+     */
+    ignoreNativeValidation?: boolean;
```
nicolechung, line 70: A bit of an open question: now that we are adding more `validation`-related arguments, should we consider having a `validation` object that's a grab-bag of all the validation arguments, something like:

```
validation: {
  validateOn: ...
  revalidateOn: ...
  ignoreNativeValidation: ...
  validate: ... 
}
```

nicolechung (approved):
Thank you for adding this so quickly!

simonihmig:
Thanks for the review, applied your suggestions!

PR #84, opened 2023-03-16
Convert addon to use template tag

simonihmig:
This converts all components of the addon itself (we only had this for the test-app before) to use template tags (gts).


ynotdraw:
YES!!!!! 🎉

NullVoxPopuli:
Glint / gts related type issues fixed here: https://github.com/CrowdStrike/ember-headless-form/pull/90
Looks like some real bugs are being revealed still!

and there are more work-arounds than I'm comfortable with in the types / imports -- I'll work towards fixing those

NullVoxPopuli:
The other reported issue has a fix PR here: https://github.com/emberjs/babel-plugin-ember-template-compilation/pull/16 (though, I don't have sufficient context and that could throw a wrench in things :sweat_smile: )

simonihmig:
> though, I don't have sufficient context and that could throw a wrench in things

Yeah, that's why I raised the issue first, without even trying to look into what needs to be fixed, as it _seemed_ intentional for _some_ reason. We'll see...

simonihmig:
Linting is fixed. But Embroider is broken, due to https://github.com/embroider-build/embroider/issues/1382

simonihmig:
> Linting is fixed. But Embroider is broken, due to https://github.com/embroider-build/embroider/issues/1382

This is fixed now with https://github.com/embroider-build/embroider/pull/1383. Therefore using a `unstable` release of Embroider now in our ember-try config, until we get the next stable release of Embroider!

packages/ember-headless-form/src/-private/components/control/input.gts
ynotdraw, line 1: I love seeing these components in a single file!  This is great.

packages/ember-headless-form/src/-private/utils.ts
```diff
@@ -28,3 +28,14 @@ export function mergeErrorRecord<
 
   return errors;
 }
+
+// this is copy pasted from https://github.com/emberjs/ember.js/blob/60d2e0cddb353aea0d6e36a72fda971010d92355/packages/%40ember/-internals/glimmer/lib/helpers/unique-id.ts
+// Unfortunately due to https://github.com/emberjs/ember.js/issues/20165 we cannot use the built-in version in template tags
```
ynotdraw, line 33: Oh! This is good to know.  Thanks for the helpful comment here

packages/ember-headless-form/src/-private/components/field.gts
```diff
@@ -279,4 +281,85 @@ export default class HeadlessFormFieldComponent<
...
+            triggerValidation=triggerValidation
+          )
+        )
+      }}
+    {{/let}}
+  </template>
 }
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars -- workaround for unknown modifier helper: https://github.com/typed-ember/glint/issues/410
+declare const modifier: any;
```
ynotdraw, line 365: Looks like a PR is up! https://github.com/typed-ember/glint/pull/553 🎉

packages/ember-headless-form/unpublished-development-types/index.d.ts
```diff
@@ -3,38 +3,13 @@
...
-> extends Helper<{
-  Args: {
-    Positional: [component: C];
-  };
-  Return: C extends keyof TemplateRegistry
-    ? TemplateRegistry[C]
-    : C extends string
-    ? ComponentLike<unknown>
-    : C;
-}> {}
```
ynotdraw, line 27: All of this cleanup is excellent 🎉

ynotdraw (approved):
🚀 This is awesome!

packages/ember-headless-form/src/-private/components/field.gts
simonihmig: Yeah, I saw that also. And it was merged some hours ago! :tada:

simonihmig:
Hey @camskene @joelamb @danwenzel, pulled you into this PR on a repo you might not have seen before, as Nicole (who together with Tony usually reviewed the stuff here) is out this week and I still need one more approval to be able to merge this! 😅. Would be nice if one of you could have a look, thanks!

packages/ember-headless-form/src/-private/components/control/checkbox.gts
```diff
@@ -43,4 +44,16 @@ export default class HeadlessFormControlCheckboxComponent extends Component<Head
   handleInput(e: Event | InputEvent): void {
     this.args.setValue((e.target as HTMLInputElement).checked);
   }
+
+  <template>
+    <input
+      name={{@name}}
+      type="checkbox"
```
danwenzel, line 51: I'm surprised to see double quotes here, as we're now (sort of) in JS, and they were single quotes in the hbs... 🤷

packages/ember-headless-form/src/-private/components/control/input.gts
```diff
@@ -1,5 +1,6 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import { on } from '@ember/modifier';
```
danwenzel, line 3: yay explicit imports! 🎉

packages/ember-headless-form/src/-private/components/control/radio-group.gts
```diff
@@ -0,0 +1,87 @@
...
+import HeadlessFormControlRadioGroupLabelComponent from './radio-group/label';
+import HeadlessFormControlRadioComponent from './radio-group/radio';
+
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import type { WithBoundArgs } from '@glint/template';
+
+export interface HeadlessFormControlRadioGroupComponentSignature {
+  Element: HTMLDivElement;
+  Args: {
+    // the following are private arguments curried by the component helper, so users will never have to use those
```
danwenzel, line 13: ah, I like this comment and the `@internal` syntax.  I ran into this with a component recently and was wondering if there was good messaging like this

packages/ember-headless-form/src/-private/components/control/select.gts
```diff
@@ -43,17 +45,38 @@ export interface HeadlessFormControlSelectComponentSignature {
...
+      id={{@fieldId}}
+      aria-invalid={{if @invalid "true"}}
+      aria-describedby={{if @invalid @errorId}}
+      ...attributes
+      {{on "input" this.handleInput}}
+    >
+      {{yield
+        (hash
+          Option=(component
+            HeadlessFormControlSelectOptionComponent selected=@value
```
danwenzel, line 76: don't need `ensure-safe-component` anymore?

danwenzel (approved):

packages/ember-headless-form/src/-private/components/control/select.gts
simonihmig: Nope! :tada:

With strict mode templates, the `{{component}}` helper is also more strict, not allowing a string based lookup anymore. This makes sense as you would have explicit imports anyway. And by that, Embroider can safely assume that whatever is passed to `{{component  myComponent}}` as the first argument is already something that has been imported _somewhere_ (and as such it is already "seen" by Embroider/webpack and added to the bundle), so it does not need to care about this being potentially unsafe anymore. I also learned about this more recently...

packages/ember-headless-form/src/-private/components/control/checkbox.gts
simonihmig: Oh, didn't realise this before. Probably our prettier config is a bit inconsistent. I did wonder why we have single quotes in hbs before (which in other projects were usually double quotes)...

PR #113, opened 2023-04-12
Tell changesets to ignore docs app

NullVoxPopuli:
Noticed on https://github.com/CrowdStrike/ember-headless-form/pull/81

that the docs app wasn't omitted from changeset


simonihmig:
Does that mean we cannot create explicit changeset entries for the docs-app anymore, that we want in the changelog? In most repos (like `release-it` based ones) we have `documentation` labelled entries also!

NullVoxPopuli:
Oh! I hadn't considered that.
The changelog entries would be in docs-app/CHANGELOG.md

If that's what you want feel free to close this pr.
I thought incruding the docs app mas a mistake 😅

Issue #109, opened 2023-03-30
The ember-headless-form-changeset example in the docs does not work


roomman:
I have followed the installation steps for this addon and `ember-changeset`/`ember-changeset-validations`. Using the example you provide I found two issues:

1) because the `validate-changeset` helper has no arguments `@validate={{validate-changeset}}` resulted in an error and required `@validate={{helper "validate-changeset"}}` instead; and

2) more crucially, I get the `Unhandled Promise Rejection: TypeError: this.args.validate is not a function. (In 'this.args.validate?.(this.internalData,Array.from(this.fields.keys()))', 'this.args.validate' is an instance of CurriedValue)` on submission of an empty form.

**Steps to replicate**
1. Clone the repo: https://github.com/roomman/ember-headless-forms-changeset-test
2. `npm install`
3. `ember s`
4. Click on the "Submit" button

This should be enough to replicate and view the error.

I hope this is helpful. 👍🏻

roomman:
Importing the helper into the backing class and exposing it via a getter works around this for me:

```ts
// test-form.ts
import Component from '@glimmer/component';
+ import { validateChangeset } from 'ember-headless-form-changeset';

export default class TestFormComponent extends Component {
  data: MyFormData = {};

+  get validateHelper() {
    return validateChangeset;
  };

  validations = {
    name: validatePresence(true),
    email: validateFormat({ type: 'email' }),
  };

  handleSubmit({ firstName, email }: MyFormData) {
    alert(`Hello ${firstName} with email ${email}!`);
  }
}
```
```hbs
// test-form.hbs
<HeadlessForm
  @data={{changeset this.data this.validations}}
  @dataMode="mutable"
  @onSubmit={{this.handleSubmit}}
  @validate={{this.validateHelper}}
  as |form|
>
  <form.Field @name="name" as |field|>
    <div class="my-2 flex flex-col">
      <field.Label>Name</field.Label>
      <field.Input class="border rounded px-2" />
      <field.Errors />
    </div>
  </form.Field>

  <form.Field @name="email" as |field|>
    <div class="my-2 flex flex-col">
      <field.Label>Email</field.Label>
      <field.Input class="border rounded px-2" />
      <field.Errors />
    </div>
  </form.Field>

  <button type="submit">Submit</button>
</HeadlessForm>
```

simonihmig:
Thanks for reporting, and sorry for the late feedback!

[...]

PR #118, opened 2023-04-18
Fix changeset helper

simonihmig:
The previous API worked by passing the helper as-is without actually invoking it: `@validate={{validate-changeset}}`, as the expected return value of the helper is a function itself. But this does not work when globally resolving the helper by its string reference, i.e. when not using `<template>` tag or Embroider. This change fixes the API of the helper, but requires your usage to change from `@validate={{validate-changeset}}` to `@validate={{(validate-changeset)}}`, invoking it without any additional arguments.

Fixes https://github.com/CrowdStrike/ember-headless-form/issues/109

### DIFF
--- a/packages/changeset/src/helpers/validate-changeset.ts
+++ b/packages/changeset/src/helpers/validate-changeset.ts
@@ -11,48 +11,45 @@ import { assert } from '@ember/debug';
  * - pass this helper into the form's `@validate` hook `@validate={{validateChangeset}}`
  * - opt-in to `@dataMode="mutable"`
  */
-const validateChangeset: FormValidateCallback<EmberChangeset> = async (
-  changeset,
-  fields
-) => {
-  assert(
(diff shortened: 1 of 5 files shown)


packages/changeset/src/helpers/validate-changeset.ts
```diff
@@ -11,48 +11,45 @@ import { assert } from '@ember/debug';
...
-    return;
-  }
-
-  const errorRecord: ErrorRecord<Record<string, unknown>> = {};
-
-  for (const { key, value, validation } of changeset.get('errors')) {
-    if (!errorRecord[key]) {
-      errorRecord[key] = [];
+export default function validateChangeset(): FormValidateCallback<EmberChangeset> {
+  return async (changeset, fields) => {
```
simonihmig, line 15: The diff here makes this change look more complex than it is: basically the helper is not the validation function itself anymore, but a function that _returns_ that validation function. Nothing else has changed here.

ynotdraw (approved):

PR #121, opened 2023-04-21
Input numbers

### LINKED ISSUE #116: Inputs dont support numbers
Thanks for the great addon!

An issue I've been running into is that inputs don't currently support numbers.

```
<HeadlessForm @data={{this.newModel}} @onSubmit={{perform this.saveModel}} as |form|>
      <form.Field @name="count" as |field|>
          <field.Label>Count</field.Label>
          <field.Input @type="number"/>
      </form.Field>
</HeadlessForm
```

If `newModel.count` has a value i.e. 2 you will get an error of `Assertion Failed: Only string values are expected for count, but you passed number`

Also, if `newModel.count` is a number prior to passing it to the form, it will throw the same error.

simonihmig:
For `@type="number"` Inputs we support passing its value as a real number, and parse and return it as a number as well (for any input, even number inputs, it's value is always a string natively).

Fixes #116

### DIFF
--- a/packages/ember-headless-form/src/-private/components/field.gts
+++ b/packages/ember-headless-form/src/-private/components/field.gts
@@ -268,7 +268,7 @@ export default class HeadlessFormFieldComponent<
 
   get valueAsStringOrNumber(): string | number | undefined {
     assert(
-      `Only string values are expected for ${String(
+      `Only string or number values are expected for ${String(
         this.args.name
       )}, but you passed ${typeof this.value}`,
       typeof this.value === 'undefined' ||
--- a/packages/ember-headless-form/src/-private/components/field.gts
+++ b/packages/ember-headless-form/src/-private/components/field.gts
@@ -266,6 +266,19 @@ export default class HeadlessFormFieldComponent<
     return this.value;
   }
 
+  get valueAsStringOrNumber(): string | number | undefined {
+    assert(
+      `Only string values are expected for ${String(
+        this.args.name
+      )}, but you passed ${typeof this.value}`,
+      typeof this.value === 'undefined' ||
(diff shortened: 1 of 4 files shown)


test-app/tests/integration/components/headless-form-data-test.gts
```diff
@@ -30,6 +30,7 @@ module('Integration Component HeadlessForm > Data', function (hooks) {
         country: 'USA',
         comments: 'lorem ipsum',
         acceptTerms: true,
+        age: 21,
```
simonihmig, line 33: I wasn't entirely sure about the correct value here, but I figured it should be just about right!? 🙃

packages/ember-headless-form/src/-private/components/field.gts
```diff
@@ -266,6 +266,19 @@ export default class HeadlessFormFieldComponent<
     return this.value;
   }
 
+  get valueAsStringOrNumber(): string | number | undefined {
+    assert(
+      `Only string values are expected for ${String(
```
ynotdraw, line 271: Should this be:

```suggestion
      `Only string or number values are expected for ${String(
```

test-app/tests/integration/components/headless-form-data-test.gts
ynotdraw: 😂 😂 😂 I wish!!! 😂 😂 😂

ynotdraw (approved):

packages/ember-headless-form/src/-private/components/field.gts
simonihmig: I actually just overlooked that! Thanks for the suggestion, will apply that!

PR #124, opened 2023-04-24
docs: Update changeset with other options of how to create a changeset

nicolechung:
## Description
Code example partly copied from: https://github.com/poteto/ember-changeset-validations/#usage

@ynotdraw  and I were pairing on implementing `ember-headless-form` and we noticed  the `changeset` is created by the parent component / page and so thought we should update the docs to mention how to handle this other case.

Just so folks know that they don't HAVE to do:
```hbs
@data={{changeset this.data this.validations}}
```


simonihmig (approved):
Approved, with some minor suggestions

danwenzel (approved):

PR #129, opened 2023-05-02
Update Glint to 1.0

simonihmig:
Besides the library updates it also tries to use the new `{{modifier}}` helper types instead of the existing hacks.

- [x] blocked on https://github.com/typed-ember/glint/issues/570


ynotdraw (approved):
🚀

Issue #130, opened 2023-05-03
Clarify semantics of updating `@data` vs. changed data


simonihmig:
> This issue has its origins in #117, some bits are copied here to focus the discussion only on the topic of what should happen when `@data` or a tracked property (`@data.foo`) is updated, while the form might have "dirty" data itself.

headless-form takes the approach of being strict about DDAU, even though in the case of forms there are some more or less valid cases for allowing two-way bindings maybe. But we don't do this, so the user will only get the new data by receiving it from `@onSubmit`. That means we have the original (unaltered) data coming from the caller as `@data`, and we have an internal data object that tracks the "dirty" data (user has filled in some fields, making those diverge from `@data`). What is displayed in the form is then basically the union of these, with any dirty value overriding the original one. Resetting the form then basically means resetting the internal data to an empty object, so all of `@data` is visible again (for forms used for "create" operations this is mostly just an empty object).

Im trying to incorporate the different cases of the state of the input (pristine - has never been touched by the user - vs. dirty - user has entered a different value than the initial one) and the action that is happening.

So the question we have to answer is: what would a user expect to happen? I.e. what should the `?` be in:

|                             | `@data.foo` is updated (`@tracked`)               | `@data` is reassigned             |
|-----------------------------|-----------------------------------------------|---------------------------------|
| input[name=foo] is pristine | ?               | ? |
| input[name=foo] is dirty    | ? | ?                               |

## 1. data gets copied initially, loses connection with `@data` afterwards

|                             | `@data.foo` is updated (`@tracked`)               | `@data` is reassigned             |
|-----------------------------|-----------------------------------------------|---------------------------------|
| input[name=foo] is pristine | input is not updated (keeps original in value)     | input is not updated (keeps original in value) |
| input[name=foo] is dirty    | input is not updated  (keeps filled in value) | input is not updated  (keeps filled in value)                             |

## 2. changed data (dirty fields) is preserved, only untouched fields reflect `@data` changes

|                             | `@data.foo` is updated (`@tracked`)               | `@data` is reassigned             |
|-----------------------------|-----------------------------------------------|---------------------------------|
| input[name=foo] is pristine | input is updated with `@data.foo`               | input is updated with `@data.foo` |
| input[name=foo] is dirty    | input is not updated  (keeps filled in value) | input is not updated  (keeps filled in value)                               |

FWIW, I think this is how `ember-changeset` behaves, having a similar concept of changed vs. original data

[...]

Issue #117, opened 2023-04-17
Issues resetting form data


averydev:
Hi there! 

I'm having an issue resetting / changing the form data values. 

My use case is a form and a list, where after you add an item via a HeadlessForm, it is added to the list, the values are cleared, and you can create a new item. 

* Setting the values of the object passed as `@data` doesn't reset the values present in the form
* setting `@data` to a new object also doesn't effect values in the form

Ideally either there would be a yielded method to change the data, or mutating the data via another approach would effect the data in the form.

simonihmig:
I agree we need to support this, will look into this!

eelke:
I noticed that resetting the form (using a simple `<input type='reset'>Reset</input>` also does not reset validation errors.

simonihmig:
> (This might be worth it's own issue?)

Yup (pun intended): #122

simonihmig:
I worked on an elaborate solution for this, until I realized it actually does not serve the use case. 🙈😅

So back to the drawing board, trying to properly design this first before implementing it...

For some additional context: headless-form takes the approach of being strict about DDAU, even though in the case of forms there are some more or less valid cases for allowing two-way bindings maybe. But we don't do this, so the user will only get the new data by receiving it from `@onSubmit`. That means we have the original (unaltered) data coming from the caller as `@data`, and we have an internal data object that tracks the "dirty" data (user has filled in some fields, making those diverge from `@data`). What is displayed in the form is then basically the union of these, with any dirty value overriding the original one. Resetting the form then basically means resetting the internal data to an empty object, so all of `@data` is visible again (for forms used for "create" operations this is mostly just an empty object).

Im trying to incorporate the different cases of the state of the input (pristine - has never been touched by the user - vs. dirty - user has entered a different value than the initial one) and the action that is happening (already taking into account #122 here in the last two columns). Here is an initial suggestion for further discussion:


|                             | `@data.foo` is updated (`@tracked`)               | `@data` is reassigned             | input[type=reset] clicked       | yielded reset action called     |
|-----------------------------|-----------------------------------------------|---------------------------------|---------------------------------|---------------------------------|
| input[name=foo] is pristine | input is updated with `@data.foo`               | input is updated with `@data.foo` | input is updated with `@data.foo` | input is updated with `@data.foo` |
| input[name=foo] is dirty    | input is not updated  (keeps filled in value) | ?                               | input is updated with `@data.foo` | input is updated with `@data.foo` |

[...]

PR #132, opened 2023-05-12
Support reactivity when `@data` is updated

simonihmig:
This supports updates of `@data` (or any of its tracked properties) getting rendered into the form, while previously filled in ("dirty") data is being preserved. This is the implementation for case `#2` of #130.

Will add some sections to the docs when #117 is also tackled, as these features are related.

### DIFF
--- a/packages/ember-headless-form/src/components/headless-form.gts
+++ b/packages/ember-headless-form/src/components/headless-form.gts
@@ -1,5 +1,5 @@
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
+import { cached, tracked } from '@glimmer/tracking';
 import { assert, warn } from '@ember/debug';
 import { hash } from '@ember/helper';
 import { on } from '@ember/modifier';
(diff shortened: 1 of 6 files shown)


packages/ember-headless-form/src/components/headless-form.gts
```diff
@@ -198,10 +198,56 @@ export default class HeadlessFormComponent<
...
-    this.args.dataMode == 'mutable' && this.args.data
-      ? this.args.data
-      : (new TrackedObject(this.args.data ?? {}) as DATA);
+  internalData: DATA = new TrackedObject({}) as DATA;
+
+  @cached
+  get effectiveData(): DATA {
+    const obj: DATA = this.args.data ?? ({} as DATA);
+
+    if (this.args.dataMode == 'mutable') {
```
ynotdraw, line 207: Super nit, but do we want strict equal here?

```suggestion
    if (this.args.dataMode === 'mutable') {
```

packages/ember-headless-form/src/components/headless-form.gts
```diff
@@ -198,10 +198,56 @@ export default class HeadlessFormComponent<
...
+      return obj;
+    }
+
+    const { internalData } = this;
+
+    return new Proxy(obj, {
+      get(target, prop) {
+        return prop in internalData
+          ? internalData[prop as keyof DATA]
+          : Reflect.get(target, prop);
```
ynotdraw, line 217: TIL about [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect)!

ynotdraw (approved):

packages/ember-headless-form/src/components/headless-form.gts
simonihmig: Oh, absolutely, good catch! 

Shouldn't ESLint have catched that though? 🤔

PR #136, opened 2023-05-15
Yield `submit` and `reset` actions

simonihmig:
`<HeadlessForm>` yields `submit` (I thought we already had this, but apparently we did not) and `reset` actions (part of #117), that can be used in place of the native buttons.

### DIFF
--- a/test-app/tests/integration/components/headless-form-reset-test.gts
+++ b/test-app/tests/integration/components/headless-form-reset-test.gts
@@ -6,6 +6,7 @@ import { module, test } from 'qunit';
 
 import { HeadlessForm } from 'ember-headless-form';
 import { setupRenderingTest } from 'test-app/tests/helpers';
+import { on } from '@ember/modifier';
 
 interface TestFormData {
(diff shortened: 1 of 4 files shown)


test-app/tests/integration/components/headless-form-reset-test.gts
```diff
@@ -153,4 +154,159 @@ module('Integration Component HeadlessForm > Reset', function (hooks) {
         .doesNotExist('form.submissionState is resetted');
     });
   });
+
+  module('reset action', function () {
```
simonihmig, line 158: This is duplicating the same tests as for the reset button. Not sure if/how we should try to DRY this, but I thought for tests some duplication is ok, instead of trying to be too clever in tests?

ynotdraw: > for tests some duplication is ok, instead of trying to be too clever in tests

100% agree!

ynotdraw (approved):

PR #134, opened 2023-05-15
Add support for `reset` button

simonihmig:
Part of #117, this adds support for resetting the state of the form using a native reset button. Yielding a `reset` action will come in a separate PR.

**Merge #132 first!**

PR #144, opened 2023-05-26
Add documentation on updating and resetting form state

simonihmig:
Add the missing docs for #132, #134, #136

### DIFF
--- a/docs/usage/data/index.md
+++ b/docs/usage/data/index.md
@@ -24,3 +24,9 @@ By default `@data` is immutable, i.e. the addon will only read from it. For hand
 But there are use cases where you would want to mutate the data immediately when the user has changed some field. This is especially the case when the data is already an object that has some "buffering" capabilities, shielding its original source data from premature mutations, as with libraries like [ember-changeset](https://github.com/poteto/ember-changeset) or [ember-buffered-proxy](https://github.com/yapplabs/ember-buffered-proxy).
 
 To do so, pass `@dataMode="mutable"` to the form component!
+
+## Updating data
+
+As we learned above, by default the form data passed as `@data` is immutable, and any changes due to the user entering or selecting new data as part of interacting with the form is stored as a separate copy, which we call "dirty" data here (as it deviates from the original data, but is not "persisted" yet). If you change `@data` (either by reassigning a whole new object or updating tracked properties), the form will automatically reflect those changes, i.e. form fields will update their value.
+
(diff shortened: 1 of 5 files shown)


docs/usage/data/index.md
```diff
@@ -24,3 +24,9 @@ By default `@data` is immutable, i.e. the addon will only read from it. For hand
...
 
 To do so, pass `@dataMode="mutable"` to the form component!
+
+## Updating data
+
+As we learned above, by default the form data passed as `@data` is immutable, and any changes due to the user entering or selecting new data as part of interacting with the form is stored as a separate copy, which we call "dirty" data here (as it deviates from the original data, but is not "persisted" yet). If you change `@data` (either by reassigning a whole new object or updating tracked properties), the form will automatically reflect those changes, i.e. form fields will update their value.
+
+However, this will _not_ happen for any non-pristine fields, i.e. fields that have new user-entered dirty data, as this would mean irrevocably losing that data. But if that is what you want, you need to explicitly wipe out that dirty data by [resetting the form](./reset). So to keep an existing form in place, but just replace all of its data, you would need to do both reassigning new `@data` and resetting the dirty state!
```
nicolechung, line 32: Total nit: I like the link to resetting here, but I would like a code snippet on either this page or the `reset` page on how to reset.

Reason being, I think folks scan until they see a code sample and likely copy-paste it.

docs/usage/reset/demo/reset.md
```diff
@@ -0,0 +1,99 @@
...
+    },
+  ];
+
+  @tracked
+  selectedIndex = 0;
+
+  resetForm;
+
+  @action
+  assignReset(reset) {
```
nicolechung, line 81: Nit: maybe there should be a comment here that this is only necessary if you want to do something with the `reset` callback?

docs/usage/reset/demo/reset.md
```diff
@@ -0,0 +1,99 @@
+# Resetting dirty form state
+
+The following kitchen-sink like example shows all three ways to reset form state:
+
+- by clicking the native reset button
```
nicolechung, line 5: nit: I think this can be split into multiple demos.

My concern is that folks with copy-paste this and have all three methods (when perhaps they just need the native reset button).

Could be
- default native reset
- reset "link"
- calling `reset` and a scenario of why this might be desired.

nicolechung (approved):

ynotdraw (approved):
🚀

docs/usage/reset/demo/reset.md
simonihmig: I currently don't have much time, so I decided to merge this as-is, but created this issue to pick up and track your suggestion: #149

PR #147, opened 2023-06-01
Upgrade ember-async-data

NullVoxPopuli:
version 0.7.0 isn't compatible with native types from ember-source.

Includes change from https://github.com/CrowdStrike/ember-headless-form/pull/148


simonihmig (commented):
LGTM, and CI seems also happy. But before merging I'd like to wait at least a day or two on feedback for [this question](https://github.com/tracked-tools/ember-async-data/pull/473/files#r1214006826)

simonihmig (approved):

PR #148, opened 2023-06-01
Noticed CI was failing because folks' (me) local pnpm is newer than v7

NullVoxPopuli:
this PR adds pnpm to the volta config and downgrades the lockfile.

This should get PRs passing again (and greatly reduce the diff of this one) https://github.com/CrowdStrike/ember-headless-form/pull/147

For fun, I'd like to show off my new CI config: https://github.com/universal-ember/ember-primitives/blob/main/.github/workflows/ci.yml#L31
It helps save the environment by taking no more than a minute and some seconds of total runtime in the event most of the tests / etc would be no-op (mostly full cache hit) -- it also uses turbo, respects volta.pnpm, etc :tada: 

Also the red status on cloudflare is likely due to me submitting from a fork -- a secure way to still have preview deploys is demonstrated here: 
- https://github.com/NullVoxPopuli/limber/blob/main/.github/workflows/ci.yml#L46
- https://github.com/NullVoxPopuli/limber/blob/main/.github/workflows/deploy-preview.yml


simonihmig (approved):
Thanks!

PR #171, opened 2023-07-23
Tells you where to import from

evoactivity:
It wasn't obvious to me where the component should be imported from. Updated the docs to make it obvious.

### DIFF
--- a/docs/typescript/index.md
+++ b/docs/typescript/index.md
@@ -10,9 +10,10 @@ This addon is written in TypeScript, and strives to give TypeScript users a firs
 > Type changes during minor upgrades may be considered breaking changes to some.
 > The goal, is to follow the advice of [Semantic Versioning for TypeScript Types](https://www.semver-ts.org/) eventually. Until noted in the Changelog that our TypeScript types fall under SemVer, consider type changes as "bugfixes" for patch releases.
 
+
 ## Usage with classic templates (hbs)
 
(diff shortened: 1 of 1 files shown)


docs/typescript/index.md
```diff
@@ -10,9 +10,10 @@ This addon is written in TypeScript, and strives to give TypeScript users a firs
...
+All the components, which are `<HeadlessForm>` as well as those [yielded from it](./usage), have strict Glint types. In case you are already using _strict mode templates_ via [ember-template-imports](https://github.com/ember-template-imports/ember-template-imports), you only need to import the component from `ember-headless-form/components/headless-form` and you are ready to go. For the more likely case of using classic `.hbs` templates, Glint requires a template registry to be set up, that declares a mapping of component _names_ to their types. This addon follows the [recommendations for Glint-enabled addons](https://typed-ember.gitbook.io/glint/using-glint/ember/using-addons#using-glint-enabled-addons), so add the addon's provided registry to your app's own registry as follows:
```
simonihmig, line 16: Thanks for working on this!

Two things here though:
* the better way to import this is straight from index, so like [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/test-app/tests/integration/components/headless-form-basic-test.gts#L6)
* This is under the "classic templates" section, which is kinda the opposite of where you would look for this information, right? I would suggest to add a separate `## Using with template tag` section here and showing a full import example as a code blcok.

evoactivity:
Sorry it took so long for me to circle back to this.

I've updated the docs with the changes requested and added an example.

PR #422, opened 2024-07-12
Use corepack instead of volta

simonihmig:
Having corepack enabled locally and volta in use is messing things up, so switching to just corepack.

@kellyselden @Exelord  I remember some discussion around this... this is the way to go for now, is it? We don't have a good and solid way to let volta manage the node version and corepack the package manager, do we?


ynotdraw (approved):

kellyselden:
This seems fine to me.

kellyselden (approved):

PR #424, opened 2024-07-15
Update ember-oss-docs and its peers

simonihmig:
Fixes #292

PR #481, opened 2024-09-27
Fix changeset bumping majors

simonihmig:
This is trying to fix the [changeset issue](https://github.com/changesets/changesets/issues/1011), which would have caused the next release (#424) to do major versions bumps for packages that have ember-headless-form as a peer dep. 

Following the same steps as https://github.com/fpapado/svg-use/pull/26.

Note: we can only validate this _after_ merging, when the changeset app has updated the release PR! 😕

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -14,5 +14,9 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["test-app"]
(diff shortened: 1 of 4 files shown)


ynotdraw (approved):
🤞 🤞 🤞

basz:
Ah I missed that comment about the release plan bumps. Did not mean to pressure.

Issue #379, opened 2024-05-22
Yield the current form data


simonihmig:
Currently, the form data as changed by the user, is only available for the `@onSubmit` callback (unless you are in mutable mode). We could yield `this.effectiveData` though to give access to that data to the template, for e.g. dynamic forms that render/unrender new form elements based on the user input.

ErvinSabic:
@simonihmig I went ahead and implemented this. If you could check out the PR I created when you get the chance.

ErvinSabic:
This can be closed now. 
@simonihmig

PR #540, opened 2025-01-13
added yielding of current form data. issue #379 + tests

ErvinSabic:
I went ahead and implemented #379 which I ended up needing myself since I have a fairly complicated form I'm using this on.

### DIFF
--- a/packages/ember-headless-form/src/components/headless-form.gts
+++ b/packages/ember-headless-form/src/components/headless-form.gts
@@ -616,7 +616,7 @@ export default class HeadlessFormComponent<
           validationState=this.validationState
           submissionState=this.submissionState
           isInvalid=this.hasValidationErrors
-          effectiveData=this.effectiveData
+          data=this.effectiveData
           rawErrors=this.visibleErrors
           submit=this.onSubmit
           reset=this.onReset
--- a/packages/ember-headless-form/src/components/headless-form.gts
+++ b/packages/ember-headless-form/src/components/headless-form.gts
@@ -616,6 +616,7 @@ export default class HeadlessFormComponent<
           validationState=this.validationState
           submissionState=this.submissionState
           isInvalid=this.hasValidationErrors
+          effectiveData=this.effectiveData
           rawErrors=this.visibleErrors
           submit=this.onSubmit
           reset=this.onReset
--- a/packages/ember-headless-form/src/components/headless-form.gts
+++ b/packages/ember-headless-form/src/components/headless-form.gts
@@ -137,7 +137,10 @@ export interface HeadlessFormComponentSignature<
          */
         reset: () => void;
 
-        data?: DATA;
+        /**
+         * The current form data that is yielded on the form itself and can be used within the template.
+         */
+        data: DATA;
       }
     ];
   };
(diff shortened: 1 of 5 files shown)


ErvinSabic:
Will fix this real quick give me like 20 min.

ErvinSabic:
Should be good now. Review when you get the chance.

ErvinSabic:
🤌 I'm not sure why that's not passing the embroiderer checks? 

I'll be able to mess with this a bit more later.

packages/ember-headless-form/src/components/headless-form.gts
```diff
@@ -136,6 +136,8 @@ export interface HeadlessFormComponentSignature<
          * Yielded action that will reset form state, same as when triggering the native `reset` event on the form.
          */
         reset: () => void;
+
+        data?: DATA;
```
simonihmig, line 140: It's been a while since I worked on this hands-on, but I wonder if this shouldn't be just `DATA`, without the `?`. `DATA` captures whatever the shape of the data is, like everything in it could be optional. But we just pass it down,  we are never _not_ passing, so I think we can just propagate the `DATA` type, I don't see when _we_ would cause it to be `undefined`.

simonihmig (changes requested):

simonihmig:
> I'm not sure why that's not passing the embroiderer checks?

Seems unrelated to your PR, I have found a case in a different build where they also failed. Need to investigate, but nothing in your PR is Embroider related, so I am confident with merging this if these are the only ones failing

Windvis:
> > I'm not sure why that's not passing the embroiderer checks?
> 
> Seems unrelated to your PR, I have found a case in a different build where they also failed. Need to investigate, but nothing in your PR is Embroider related, so I am confident with merging this if these are the only ones failing

I was intrigued so took a look at the code :smile: 

It seems the ember-try config is still [installing an old, unstable version](https://github.com/CrowdStrike/ember-headless-form/blob/ad45c1aada376ffef7afee9922cd7a7e317d03cf/test-app/config/ember-try.js#L58-L60) of the Embroider packages. That combined with ember-source 6.1 being a v2 addon now might be the cause since those old versions most likely don't know how to deal with that yet.

Looking at the logs, it seems that the [ember-source override to 4.8](https://github.com/CrowdStrike/ember-headless-form/blob/ad45c1aada376ffef7afee9922cd7a7e317d03cf/test-app/config/ember-try.js#L56) also doesn't work and the release version is still being used there as well, which could explain why those also fail now.

I think there is a bug in the `@embroider/test-setup` package where the [scenarios' devDeps object](https://github.com/embroider-build/embroider/blob/f7790111440d0e7615cd94067aca96e5653ea41e/packages/test-setup/src/index.ts#L12-L18) is unintentionally shared, and then [modified ](https://github.com/embroider-build/embroider/blob/f7790111440d0e7615cd94067aca96e5653ea41e/packages/test-setup/src/index.ts#L107)when it is being merged with the one provided by the app.

packages/ember-headless-form/src/components/headless-form.gts
ErvinSabic: I'll try out a few different use cases just to get a feel for what a developer might expect when using this.

ErvinSabic:
Should be good to go.

simonihmig (approved):
Thank you!

Issue #483, opened 2024-09-30
Number input does not work for all locales


simonihmig:
The input control with type=number calls `parseFloat()` to return a proper number as the form value. But this does not take different locals into account, e.g. those that use a comma as the decimal separator.

We could make use of e.g. https://github.com/ApelegHQ/intl-number-parser or some equivalent npm package or bring our own implementation.

ErvinSabic:
So about this @simonihmig 

`<input type="number" />` is bad at supporting this kind of localization. From my testing by changing my location through chrome dev tools, if I set my country to say Germany, inputting `"1.234,56"` is not allowed. 

From my testing regardless of the user's locale, they need to input `"1234.56"`. 

My initial idea was to implement this by changing the number input into a text input and handling the formatting on a "focus out" event. The issue with this is the data is changed by the focus out event, then changed in the text input as well. 

In my instance, if I input `"1.234,56"` then focus out, inside of the text box it gets turned into `"1234.56"`, great. That's how the data I presume is expected to be saved in terms of the data. This is something I understand as a *developer*. 

*However,* if I, as a confused user go to click on the box again (because my number changed, what the heck?!) and unclick it, my `"1234.56"` is now turned into `"123456"`, which is a far cry from the intended number. I can probably handle this quickly 

*Now* if we *really* want to implement this, we're going to need to take it a step further just for the number inputs to keep a version of the data that is the "saved number" that is used as the actual data for the form and the "formatted number" that is shown to the user in text box. We can also probably just handle the parsing on submission or during the validation step. 

**But I had a thought:** 

As an American who has absolutely 0 idea on how people in other parts of the world use number inputs, are they used to following the US decimal system (1234.56) when needing to input these numbers? Or is there the expectation that their locale is supported? 

I can implement a solution but it may come across more complicated than I might like for something like this. **Just** for those pesky number inputs! 

If people are used to using number inputs as they've always been on the internet (only supporting numbers and "." as far as I can tell), they may find it confusing when they're suddenly expected to use their locale which may cause a mis-input. 

I'll be asking some people around the world what their expectation is when they go to fill out forms online and report back. But I wanted to hear your thoughts on this.

ErvinSabic:
I might just make a whole other localized number input and just leave the number type alone so it stays consistent with the rest of the library.

simonihmig:
> From my testing regardless of the user's locale, they need to input "1234.56"

[...]

Issue #468, opened 2024-09-13
Entering  "2.01" in a number in an input is impossible.


averydev:
With an input with a type of "number", there's a problem. 

To replicate create an input like so:

```
<field.Input @type="number" step="0.01" />
```
When you try and type in `2.01` after typing the `0` the value will be reset to `2` with no decimal point. You can copy-paste the value in just fine. 

What's happening is in `handleInput(e: Event | InputEvent): void`  the input is getting transformed with `parseFloat(e.target.value)`. That's the right thing (basically) because otherwise number fields would return string values. 

However JS doesn't have the concept of "2.0" so it is parsed into "2" and set, wiping out the decimal point. 

In short, it's impossible to type in decimal values that contain 0. 

I've added a test replicating the issue here:
https://github.com/stackdevelopment/ember-headless-form-number-issue/tree/number-input-issue

I'd be happy to fix it if you point me in the direction for how you'd like to approach it.

simonihmig:
Thanks for reporting @averydev , and sorry for the late response!

Great that you have already a nice reproduction of this as a test! Happy to accept this as a PR alone, even when not coming together with the actual fix! 

If you want to work on a fix, my initial thinking was to add some logic where the `setValue` is called [here](https://github.com/CrowdStrike/ember-headless-form/blob/main/packages/ember-headless-form/src/-private/components/control/input.gts#L110) to _not_ call `setValue` when the input has a (string) value that represents the same number, but would cause `parseFloat` to strip something out. Thinking aloud:
* check if the (string) value end on `.`, or has `.` and ends on `0`?
* check if `value !== String(parseFloat(value))`

Does that make sense?

simonihmig:
Looking at the code around `parseFloat`, I  just realized there is even another issue with that: https://github.com/CrowdStrike/ember-headless-form/issues/483 🤦

ErvinSabic:
I'll pick this up. @simonihmig

PR #541, opened 2025-01-18
WIP: Locale Inputs, Resolving broken embroiderer dependencies, Unfocus Formatting

ErvinSabic:
This seeks to solve a few issues: 

#468 
#483

Dependency issue mentioned by Windvis here - https://github.com/CrowdStrike/ember-headless-form/pull/540#issuecomment-2592727387

Will mark as ready when I believe it can be submitted for review.

### DIFF
--- a/packages/ember-headless-form/src/-private/components/control/input.gts
+++ b/packages/ember-headless-form/src/-private/components/control/input.gts
@@ -50,6 +50,14 @@ export interface HeadlessFormControlInputComponentSignature {
      */
     type?: InputType;
 
+    /**
+     * Some inputs are specific to locales, like numbers. This
+     *
+     *
+     *
+     */
+    locale?: string
+
     // the following are private arguments curried by the component helper, so users will never have to use those
 
     /*
@@ -106,10 +114,18 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
   @action
   handleInput(e: Event | InputEvent): void {
     assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
-    this.args.setValue(
-      this.type === 'number' ? parseFloat(e.target.value) : e.target.value
-    );
+    this.args.setValue(e.target.value);
   }
+
+  @action
+  handleUnfocus(e: Event | InputEvent): void {
+    assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
+
+    if(this.type === "number"){
+      this.args.setValue(parseFloat(e.target.value));
+    }
+  }
+
   <template>
     <input
       name={{@name}}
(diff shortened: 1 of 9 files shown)


simonihmig:
Hey @ErvinSabic, thanks for working on this! Just kicked off CI. Tests including Embroider are passing! :tada: But some type errors, fyi

ErvinSabic:
@simonihmig I encourage you to give this a try locally. I'm going to be writing more tests for it and I want to try it on a mobile device as well to see what happens with the caret positioning. 

It should work for:
Whole numbers,
Decimals, 
Currencies,
Percentages, 
Units. 

Does not support:
Scientific Notation
Compact Notation

packages/ember-headless-form/src/-private/components/control/input.gts
```diff
@@ -106,10 +105,18 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
   @action
   handleInput(e: Event | InputEvent): void {
     assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
-    this.args.setValue(
-      this.type === 'number' ? parseFloat(e.target.value) : e.target.value
-    );
+    this.args.setValue(e.target.value);
```
simonihmig, line 108: Hm, I don't think we can do that. This would mean when inputting, the value is set as a string, but when focussing out as a number. The user might not see the difference in the UI, but the internal value would be flip-flopping between string and number. This could have effects on the validation (if used). Like when you have a custom validator or using `yup` for example, the validation result might be different based on whether the  value is a string or a number!

simonihmig (commented):
@ErvinSabic thanks a lot for your effort, and sorry again for the late feedback!

I think it would really help if we could split this PR into the three different issues/features it tackles!

1. the dependency updates are really good and uncontroversial, so would be nice to get them in asap
2. for the number input issue #468, I left a comment here
3. the new locale-number component seems quite massive, I would need to play with that locally. It seems to do much more than fixing #483, like adding unit support, right? This is interesting, but really a completely new feature. Even just for changelog generation, it would help to have this extracted. On the #483 issue though, I left some comments there. There seems to be a misunderstanding as the native number input does correctly handle localized numbers (at least for me), and you cannot opt-out of localized rendering. On my machine it will always render with a comma as the decimal separator. But also as mentioned there, that assumed `parseFloat` issue might not even be an issue.

packages/ember-headless-form/src/-private/components/control/input.gts
ErvinSabic: Wouldn't parsing as a number in here cause the original issue here (https://github.com/CrowdStrike/ember-headless-form/issues/468) because the user isn't done typing?

ErvinSabic: Gotcha. Let me break this PR up into their respective issues and work on it further.

ErvinSabic:
Closing this PR as I've broken the changes up into their respective PRs - 

https://github.com/CrowdStrike/ember-headless-form/pull/550
https://github.com/CrowdStrike/ember-headless-form/pull/551
https://github.com/CrowdStrike/ember-headless-form/pull/552

PR #551, opened 2025-04-22
Resolution of issues #468 and possibly #483

ErvinSabic:
Used some of @simonihmig 's suggestions along with accounted for NaN or blank values defaulting to zero.


packages/ember-headless-form/src/-private/components/control/input.gts
```diff
@@ -120,6 +139,7 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
       aria-describedby={{if @invalid @errorId}}
       ...attributes
       {{on "input" this.handleInput}}
+      {{!-- {{on "focusout" this.handleFocusOut}} --}}
```
simonihmig, line 142: Can be removed?

packages/ember-headless-form/src/-private/components/control/input.gts
```diff
@@ -106,10 +105,30 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
   @action
   handleInput(e: Event | InputEvent): void {
     assert('Expected HTMLInputElement', e.target instanceof HTMLInputElement);
-    this.args.setValue(
-      this.type === 'number' ? parseFloat(e.target.value) : e.target.value
-    );
+
+    if(this.args.type === "number"){
```
simonihmig, line 109: Seems prettier is not checking files correctly, will need to investigate. This should have spaces, like `
```suggestion
    if (this.args.type === "number") {
```

Will not bike-shed on the PR, given that CI is not catching this as it should. I can fix that later...

packages/ember-headless-form/src/-private/components/control/input.gts
```diff
@@ -106,10 +105,30 @@ export default class HeadlessFormControlInputComponent extends Component<Headles
...
+      const valueAsNumber = parseFloat(e.target.value);
+
+      if(valueAsNumber !== this.args.value && !isNaN(valueAsNumber)){
+        this.args.setValue(valueAsNumber);
+      }
+      // This is here to avoid setting the value as NaN, instead using Zero.
+      else if(e.target.value === "" || isNaN(valueAsNumber)){
+        this.args.setValue(0);
```
simonihmig, line 117: I am not quite sure about this one. Can you explain your rationale for this case? 

Wouldn't not setting the value here be more appropriate?

simonihmig (commented):
Looking good, thanks @ErvinSabic! But we would need some test for this. @averydev had one already [here](https://github.com/CrowdStrike/ember-headless-form/compare/main...stackdevelopment:ember-headless-form-number-issue:number-input-issue), linked in #468. 

Also just merged #550, so if you could rebase this should give us a green build (ignoring the known docs deployment issue)

packages/ember-headless-form/src/-private/components/control/input.gts
ErvinSabic: done.

packages/ember-headless-form/src/-private/components/control/input.gts
ErvinSabic: Here's a video explaining my thinking - 

https://github.com/user-attachments/assets/83931e39-9b7c-48d8-b2a5-9d7e49d2553b

ErvinSabic: Defaulting to a zero is simpler than maintaining the last valid state like I do on my intl number input. Number inputs should *always* be numbers.

packages/ember-headless-form/src/-private/components/control/input.gts
ErvinSabic: I'll fix this where I see it.

ErvinSabic:
I just went ahead and pulled the latest from the base branch, added a variation of the test you'd given along with an extra.  Should test and merge in green now.

ErvinSabic:
Something seems to be going on with prettier in the linting process?

PR #552, opened 2025-04-23
Intl Number Input Feature

ErvinSabic:
This PR Consists of a input that supports the majority of the options offered by [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) by specifying your locale and the options you would like to use. The majority of the which are support (Except for Scientific Notation and Compact Notation). 

In short: 
- This is a text input emulating a number input. 
- It shows the number in the user's own locale using the specified options as they're type. (thousands separators, decimal mark, minus sign, currency/unit/percent affixes, non‑Latin numerals).
- It will support most of the Intl.NumberFormat options. Save for some edge cases (Ex, typing "2" in a input setup for units, in arabic, for liters might cause unexpected behavior).  
- It stores the canonical dot based value through the usual setValue method. Meaning whatever the user types into the input is eventually turned back into a workable number that you can store in a database. Ex: While the text input may show `"€ 1,234.56"` the data value of the input is `1234.56`. When working with percentages, it may show `"75%"` but the data value is `0.75`. 
- It rejects or adapts to inputs such as multiple decimal places, too many fraction digits, NaNs, etc. Usually by reverting to a previously valid value.
- The position of the user's input (caret) will adjust based on formatting and inputs.

### DIFF
--- a/docs/usage/controls.md
+++ b/docs/usage/controls.md
@@ -159,3 +159,28 @@ Renders a native `<select>` dropdown. It will yield another `Option` component f
   <button type='submit'>Submit</button>
 </HeadlessForm>
 ```
+
+## Localized Number
+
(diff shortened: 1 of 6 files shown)


ErvinSabic:
I'll fix these test failures.

